PR #30, opened 2023-03-31
Fixing bad breaking bug in endpoint url sanitization

robotrapta:
A funny-to-me set of coincidences let this bug through all our tests.  The endpoint sanitizer was well tested in unit tests, but the integration wasn't quite right, and the integration tests skipped the sanitizer.  That's fixed now.  As is the bug.

### DIFF
--- a/src/groundlight/client.py
+++ b/src/groundlight/client.py
@@ -43,7 +43,7 @@ class Groundlight:
     POLLING_EXPONENTIAL_BACKOFF = 1.3  # This still has the nice backoff property that the max number of requests
     # is O(log(time)), but with 1.3 the guarantee is that the call will return no more than 30% late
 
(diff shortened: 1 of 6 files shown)


src/groundlight/client.py
```diff
@@ -43,15 +43,15 @@ class Groundlight:
...
+    def __init__(self, endpoint: Optional[str] = None, api_token: str = None):
         """
         :param endpoint: optionally specify a different endpoint
         :param api_token: use this API token for your API calls. If unset, fallback to the
             environment variable "GROUNDLIGHT_API_TOKEN".
         """
         # Specify the endpoint
         self.endpoint = sanitize_endpoint_url(endpoint)
-        configuration = Configuration(host=endpoint)
+        configuration = Configuration(host=self.endpoint)
```
robotrapta, line 54: The missing `self.` was actually the bug.  The rest of this PR is cleaning up tests and such.

PR #31, opened 2023-04-01
Pil image support.  Default WAIT

### LINKED ISSUE #22: Add ability to send PIL image directly
Don't require PIL to be installed to do this.

robotrapta:
You can now submit imagequeries with PIL images.  But obviously only if PIL is installed.  Fixes #22 

Also you can set `gl.DEFAULT_WAIT` which overrides the default 30 for all submit-image-query calls.

### DIFF
--- a/src/groundlight/client.py
+++ b/src/groundlight/client.py
@@ -135,7 +135,7 @@ class Groundlight:
           - byte array or BytesIO or BufferedReader with jpeg bytes
           - numpy array of dimensions (3,W,H) in the 0-255 range in RGB order
           - PIL Image
-          Any binary format must be JPEG-encoded already.  Any pixel format will get 
+          Any binary format must be JPEG-encoded already.  Any pixel format will get
           converted to JPEG at high quality before sending to service.
         :param wait: How long to wait (in seconds) for a confident answer
         """
--- a/src/groundlight/client.py
+++ b/src/groundlight/client.py
@@ -38,6 +38,8 @@ class Groundlight:
     ```
     """
 
+    DEFAULT_WAIT = 30
+
     BEFORE_POLLING_DELAY = 3.0  # Expected minimum time for a label to post
     POLLING_INITIAL_DELAY = 0.5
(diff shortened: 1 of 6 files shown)


obj-imp (commented):
cool! we should update our docs!

does the ndarray type detection work if you pass an OpenCV frame or do you have to cast it ?

src/groundlight/client.py
```diff
@@ -123,37 +125,24 @@ def submit_image_query(
         self,
         detector: Union[Detector, str],
         image: Union[str, bytes, BytesIO, BufferedReader, np.ndarray],
```
tyler-romero, line 127: nit: add `Image.Image` to the supported input types

src/groundlight/images.py
```diff
@@ -21,8 +22,36 @@ def buffer_from_jpeg_file(image_filename: str) -> io.BufferedReader:
...
-        buf = io.BytesIO()
+    with BytesIO() as buf:
+        buf = BytesIO()
         pilim.save(buf, "jpeg", quality=jpeg_quality)
         out = buf.getvalue()
         return out
+
+
+def parse_supported_image_types(
+    image: Union[str, bytes, BytesIO, BufferedReader, np.ndarray], jpeg_quality: int = 95
```
tyler-romero, line 33: nit: add `Image.Image` to the list of supported types here as well

tyler-romero (approved):
LGTM, just a couple nits

src/groundlight/client.py
robotrapta: Good catch!

robotrapta:
> does the ndarray type detection work if you pass an OpenCV frame or do you have to cast it ?

Unfortunately, I think the answer is that it "almost works".  I think what will happen is that we will accept the image, but the colors will be wrong because OpenCV uses BGR order for color-channel, but we expect RGB.

This is certainly not awesome.  If we're getting an image as a numpy array, it's very likely it came from OpenCV.  Maybe we should require images to be BGR.  That change would be backwards incompatible though.  Filed #34 to track this.

PR #32, opened 2023-04-01
Try mkdocs for our documentation website

mjvogelsong:
Check it out: https://groundlight.github.io/python-sdk.

We can change the URL to something like `docs.groundlight.ai` when we're ready.

Caveat: We need Python 3.8+ for the doc dependencies... Although we could make a separate `pyproject.toml` inside `docs/` -- a little hacky, but reasonable if we still want to support 3.7 on the main SDK.


mjvogelsong:
Docusaurus is better 😀

PR #33, opened 2023-04-02
Try docusaurus for docs

mjvogelsong:
Check it out here: https://groundlight.github.io/python-sdk/docs/getting-started

We now have a github action that will build and deploy the site to github pages. We can change the URL to something like `docs.groundlight.ai` when we're ready. And maybe host it on Netlify or Vercel.


.github/workflows/docs.yaml
```diff
@@ -0,0 +1,47 @@
+name: docs
```
robotrapta, line 1:
```suggestion
name: docs
# Builds the documentation website and publishes it to https://groundlight.github.io
```

As we rely more on GHA, we should get in the habit of writing docstrings at the top for what they do.

docs/README.md
```diff
@@ -0,0 +1,41 @@
+# Website
+
+This website is built using [Docusaurus 2](https://docusaurus.io/), a modern static website generator.
+
```
robotrapta, line 4: Of course we'll want to update all this sample stuff.

docs/blog/2023-04-01-mdx-blog-post.mdx
robotrapta, line 1: I don't want to have a "blog" right now.  We should build a place to put blog-like articles, but I don't want them to be presented that way.

robotrapta (approved):
Great!

PR #35, opened 2023-04-03
Run tests against the python code in our documentation

mjvogelsong:
We can test the code block examples in our docs using [pytest-markdown-docs](https://github.com/modal-labs/pytest-markdown-docs)! This should help us make sure our documentation examples are always runnable.

One open question is whether it is better to make most/every example a complete example? As a user of SDK docs, I kinda prefer that, because then every code block becomes fully-functional without any assumptions. We can highlight the "important" part of the code. But I do also think there's value in having shorter code snippets to focus on the important part. There is a `python continuation` fence that we can use to test against a series of python blocks.

TODOs:
- It already found some bugs -- e.g., we don't expose a `confidence` parameter in `get_or_create_detector()`. Gotta fix that.
- It would be nice to put `asserts` inline with the code blocks, but not render them...
- Switch all tests to hit prod instead of integ


docs/docs/getting-started/managing-confidence.md
```diff
@@ -6,27 +6,34 @@ The desired confidence level is set as the escalation threshold on your detector
 
 For example, say you want to set your desired confidence level to 0.95, but that you're willing to wait up to 60 seconds to get a confident response.
 
-```python
+<!-- TODO: turn on `confidence` parameter, and re-enable tests! -->
```
mjvogelsong, line 9: Gotta fix this

tyler-romero (approved):
LGTM!

PR #37, opened 2023-04-13
Add confidence threshold to sdk

mjvogelsong:
We had `confidence` in our docs, but it wasn't actually wired up yet. So I added it here.

### DIFF
--- a/docs/docs/getting-started/managing-confidence.md
+++ b/docs/docs/getting-started/managing-confidence.md
@@ -6,7 +6,7 @@ The desired confidence level is set as the escalation threshold on your detector
 
 For example, say you want to set your desired confidence level to 0.95, but that you're willing to wait up to 60 seconds to get a confident response.
 
-<!-- TODO: turn on `confidence` parameter, and re-enable tests! -->
+<!-- We skip tests here because the tests are too slow -->
 
 ```python notest
 from groundlight import Groundlight
@@ -22,7 +22,7 @@ image = Image.open(
 d = gl.get_or_create_detector(name="trash", query="Is the trash can full?", confidence=0.95)
 
 # This will wait until either 60 seconds have passed or the confidence reaches 0.95
-image_query = gl.submit_image_query(detector=d, image=jpeg_img, wait=60)
+image_query = gl.submit_image_query(detector=d, image=image, wait=60)
 # highlight-end
 
 print(f"The answer is {image_query.result}")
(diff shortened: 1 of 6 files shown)


docs/docs/getting-started/managing-confidence.md
```diff
@@ -6,7 +6,7 @@ The desired confidence level is set as the escalation threshold on your detector
 
 For example, say you want to set your desired confidence level to 0.95, but that you're willing to wait up to 60 seconds to get a confident response.
 
-<!-- TODO: turn on `confidence` parameter, and re-enable tests! -->
+<!-- We skip tests here because the tests may be slow -->
```
mjvogelsong, line 9: I did run this doc test, but I'm not sure we want to run it in our test matrix (which would run this 20x).

src/groundlight/client.py
```diff
@@ -95,24 +95,37 @@ def list_detectors(self, page: int = 1, page_size: int = 10) -> PaginatedDetecto
         obj = self.detectors_api.list_detectors(page=page, page_size=page_size)
         return PaginatedDetectorList.parse_obj(obj.to_dict())
 
-    def create_detector(self, name: str, query: str, config_name: str = None) -> Detector:
-        obj = self.detectors_api.create_detector(DetectorCreationInput(name=name, query=query, config_name=config_name))
+    def create_detector(
+        self, name: str, query: str, *, confidence: float = DEFAULT_CONFIDENCE_THRESHOLD, config_name: str = None
```
mjvogelsong, line 99: Should this be called `confidence` or `confidence_threshold`?

robotrapta: Threshold makes sense to me - it's more clear.

docs/docs/getting-started/managing-confidence.md
robotrapta: We should think about enabling this on at least some of the platforms.  Like maybe just 3.10.

robotrapta: Actually it would make sense to only run some of the tests when PIL is installed, per earlier discussion.

src/groundlight/client.py
```diff
@@ -95,24 +95,37 @@ def list_detectors(self, page: int = 1, page_size: int = 10) -> PaginatedDetecto
...
                 raise ValueError(
-                    f"Found existing detector with name={name} (id={existing_detector.id}) but the queries don't match"
+                    f"Found existing detector with name={name} (id={existing_detector.id}) but the confidence thresholds don't match. The existing confidence threshold is {existing_detector.confidence_threshold}."
                 )
+            else:
+                return existing_detector
 
-        return self.create_detector(name, query, config_name)
+        if confidence is None:
+            confidence = self.DEFAULT_CONFIDENCE_THRESHOLD
```
robotrapta, line 127: Is it better to have the default in the SDK or in the cloud?

robotrapta (approved):
I think the SDK parameter would be better as `confidence_threshold` instead of just `confidence` but otherwise looks good.

We should thinking about making the test matrix more sophisticated.

docs/docs/getting-started/managing-confidence.md
mjvogelsong: I made some improvements in https://github.com/groundlight/python-sdk/pull/40

src/groundlight/client.py
mjvogelsong: I think it's better in the cloud - less duplication. I'll change it to default `None`, so the cloud holds the default.

PR #39, opened 2023-04-14
Standardized linting and formatting

mjvogelsong:
The package should be passing all linter and formatting checks now! I also made a few improvements on makefiles, readmes, and documentation.


.pylintrc
```diff
@@ -3,13 +3,7 @@
...
 # be loaded. Extensions are loading into the active Python interpreter and may
 # run arbitrary code.
-extension-pkg-allow-list=
-
-# A comma-separated list of package or module names from where C extensions may
-# be loaded. Extensions are loading into the active Python interpreter and may
-# run arbitrary code. (This is an alternative name to extension-pkg-allow-list
-# for backward compatibility.)
-extension-pkg-whitelist=
+extension-pkg-allow-list=pydantic
```
tyler-romero, line 6: Yay!

.pylintrc
```diff
@@ -33,7 +27,7 @@ ignore-patterns=
 # Python code to execute, usually for sys.path manipulation such as
 # pygtk.require().
 # This is a workaround until pysrc is its own package!
```
tyler-romero, line 29:
```suggestion
```

nit: remove old comment

.pylintrc
```diff
@@ -53,7 +47,7 @@ persistent=yes
 
 # Min Python version to use for version dependend checks. Will default to the
 # version used to run pylint.
-py-version=3.9
+py-version=3.7
```
tyler-romero, line 50: Does this mean we shouldnt use new python language features?

mjvogelsong: Unfortunately yeah - with the SDK, we want to support as many customers as possible, so we have to stay more backwards-compatible with a broader range of python versions and other python dependencies than with our backend stuff.

On the other hand, I do think we might want limit to 3.8+. 3.7 is end-of-life and will not get security updates afetr June 2023 (https://endoflife.date/python).

PR #41, opened 2023-04-14
Fix tests (improve `get_detector_by_name()`)

mjvogelsong:
Our tests have been failing because `get_detector_by_name()` was doing the name-filtering on the client side. This change moves the filter to the server side, so it's fast and simpler.


tyler-romero (approved):
Nice!

PR #46, opened 2023-04-18
Bugfix for `get_or_create_detector()` unique name

mjvogelsong:
I found a bug where `get_or_create_detector()` wasn't handling new detectors properly. We hadn't noticed before because we didn't have a test that operated on an explicitly unique detector name.

Longer term: maybe the "get-or-create" functionality should be happening server-side anyway?


tyler-romero (approved):
LGTM!

PR #50, opened 2023-04-20
(0.8.0rc1) Return YES/NO instead of PASS/FAIL

### LINKED ISSUE #7: ImageQuery results say PASS/FAIL not YES / NO
When I look at my ImageQuery response object, for one it's this super complicated thing (`ResultTypeEnum.binary_classification`?), but when I parse through it all, it says "FAIL" which isn't how I set up the question.  The question is set up as Answer 1: Yes, Answer 2: No.

```
ImageQuery(id='chk_2CMxGn4m6Ld8nTM6ADw7LEYVib8', type=<ImageQueryTypeEnum.image_query: 'image_query'>, created_at=datetime.datetime(2022, 7, 23, 18, 29, 21, 347772, tzinfo=tzoffset(None, -25200)), query='Is there a n03888257?', detector_id='det_2CMxEkcnQodh2QwEVonUigM1g2w', result_type=<ResultTypeEnum.binary_classification: 'binary_classification'>, result=ClassificationResult(confidence=None, label='FAIL'))
```

mjvogelsong:
The SDK should not be returning the confusing `PASS` or `FAIL` labels anymore.

I'm not a big fan of the mapper in the client, but I think it's an important fix for clarity in the short term.

Fixes #7

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ packages = [
     {include = "**/*.py", from = "src"},
 ]
 readme = "README.md"
-version = "0.7.8"
+version = "0.8.0"
 
 [tool.poetry.dependencies]
 certifi = "^2021.10.8"
--- a/src/groundlight/binary_labels.py
+++ b/src/groundlight/binary_labels.py
@@ -1,4 +1,4 @@
-"""Defines the possible values for binary class labels like Yes/No or PASS/FAIL.
+"""Defines the possible values for binary class labels like YES/NO or PASS/FAIL.
 Provides methods to convert between them.
 
 This part of the API is kinda ugly right now.  So we'll encapsulate the ugliness in one place.
@@ -11,6 +11,13 @@ from model import Detector, ImageQuery
 logger = logging.getLogger("groundlight.sdk")
 
 
+PASS_DEPRECATED = "PASS"
+FAIL_DEPRECATED = "FAIL"
+YES = "YES"
+NO = "NO"
+DEPRECATED_LABEL_NAMES = {PASS_DEPRECATED, FAIL_DEPRECATED}
+
+
 def internal_labels_for_detector(
     context: Union[ImageQuery, Detector, str],  # pylint: disable=unused-argument
 ) -> List[str]:
@@ -19,7 +26,7 @@ def internal_labels_for_detector(
     :param context: Can be an ImageQuery, a Detector, or a string-id for one of them.
     """
     # NOTE: At some point this will need to be an API call, because these will be defined per-detector
-    return ["PASS", "FAIL"]
+    return [PASS_DEPRECATED, FAIL_DEPRECATED]
 
 
 def convert_internal_label_to_display(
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -9,7 +9,7 @@ packages = [
     {include = "**/*.py", from = "src"},
 ]
 readme = "README.md"
-version = "0.7.8"
+version = "0.8.0"
```
mjvogelsong, line 12: Application code that relied on the old `PASS/FAIL` `result.label` will break! Technically, it should be a major version update, but we're not ready for 1.0.0 yet.

src/groundlight/binary_labels.py
```diff
@@ -28,11 +35,11 @@ def convert_internal_label_to_display(
 ) -> str:
     # NOTE: Someday we will do nothing here, when the server provides properly named classes.
     upper = label.upper()
-    if upper == "PASS":
-        return "YES"
-    if upper == "FAIL":
-        return "NO"
-    if upper in ["YES", "NO"]:
+    if upper == PASS_DEPRECATED:
```
tyler-romero, line 38: Is this right? I thought that 0 == Pass == No and 1 == FAIL == Yes

robotrapta:
Before we release 0.8 I’d like to address the BGR / RGB issue since that’s also a breaking change.

src/groundlight/binary_labels.py
mjvogelsong: Based on the way the prior code was set up (and the diagram I wrote to myself because I keep forgetting and this mapping is terrible), I think it's correct:

0 == PASS == YES == Normal
1 == FAIL == NO == Anomalous

IMG_1588

mjvogelsong: Also, https://github.com/positronix-ai/zuuul/blob/main/janzu/apparati/views.py#L1178 says PASS == YES and FAIL == NO

test/integration/test_groundlight.py
```diff
@@ -134,6 +135,14 @@ def test_submit_image_query_blocking(gl: Groundlight, detector: Detector):
     _image_query = gl.submit_image_query(detector=detector.id, image="test/assets/dog.jpeg", wait=2)
     assert str(_image_query)
     assert isinstance(_image_query, ImageQuery)
+    assert _image_query.result.label not in DEPRECATED_LABEL_NAMES
+
+
+def test_submit_image_query_returns_yes(gl: Groundlight):
+    # We use the "never-review" model to guarantee a "yes" answer.
+    detector = gl.get_or_create_detector(name="Always a dog", query="Is there a dog?", config_name="never-review")
+    image_query = gl.submit_image_query(detector=detector, image="test/assets/dog.jpeg", wait=5)
```
robotrapta, line 144: I like this dog.  :)  https://github.com/groundlight/python-sdk/blob/main/docs/static/img/michonne.jpg

robotrapta (approved):
Great!

PR #51, opened 2023-04-20
(0.8.0rc1) Only allow "YES"/"NO" in `add_label()`

mjvogelsong:
`add_label()` now only accepts "YES" or "NO" (case-insensitive, string or enum). If the user provides something else, they get a `ValueError`.

When reading `ImageQuery.result.label`, the user will see uppercase "YES", "NO", or "UNSURE".

### DIFF
--- a/src/groundlight/binary_labels.py
+++ b/src/groundlight/binary_labels.py
@@ -4,29 +4,30 @@ Provides methods to convert between them.
 This part of the API is kinda ugly right now.  So we'll encapsulate the ugliness in one place.
 """
 import logging
-from typing import List, Union
+from enum import Enum
+from typing import Any, Union
 
-from model import Detector, ImageQuery
+from model import ClassificationResult, Detector, ImageQuery
 
-logger = logging.getLogger("groundlight.sdk")
-
-
-PASS_DEPRECATED = "PASS"
-FAIL_DEPRECATED = "FAIL"
-YES = "YES"
-NO = "NO"
-DEPRECATED_LABEL_NAMES = {PASS_DEPRECATED, FAIL_DEPRECATED}
+logger = logging.getLogger(__name__)
 
 
-def internal_labels_for_detector(
-    context: Union[ImageQuery, Detector, str],  # pylint: disable=unused-argument
-) -> List[str]:
-    """Returns an ordered list of class labels as strings.
-    These are the versions of labels that the API demands.
-    :param context: Can be an ImageQuery, a Detector, or a string-id for one of them.
-    """
-    # NOTE: At some point this will need to be an API call, because these will be defined per-detector
-    return [PASS_DEPRECATED, FAIL_DEPRECATED]
+class Label(str, Enum):
+    YES = "YES"
+    NO = "NO"
+    UNSURE = "UNSURE"
+
+
+VALID_DISPLAY_LABELS = {Label.YES, Label.NO, Label.UNSURE}
+
+
+class DeprecatedLabel(str, Enum):
+    PASS = "PASS"
+    FAIL = "FAIL"
+    NEEDS_REVIEW = "NEEDS_REVIEW"
+
+
+DEPRECATED_LABEL_NAMES = {DeprecatedLabel.PASS, DeprecatedLabel.FAIL, DeprecatedLabel.NEEDS_REVIEW}
 
 
 def convert_internal_label_to_display(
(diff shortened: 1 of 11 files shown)


src/groundlight/binary_labels.py
```diff
@@ -1,51 +1,98 @@
...
-        return label
+    if upper in {Label.YES, DeprecatedLabel.PASS}:
+        return Label.YES.value
+    if upper in {Label.NO, DeprecatedLabel.FAIL}:
+        return Label.NO.value
+    if upper in {Label.UNSURE, DeprecatedLabel.NEEDS_REVIEW}:
+        return Label.UNSURE.value
+
+    # Safety versus extensibility: should it be an error if we get an unrecognized internal label?
     logger.warning(f"Unrecognized internal label {label} - leaving alone.")
```
mjvogelsong, line 54: In theory, this shouldn't happen because we don't support custom labels. But if it does (e.g., we push a change to the API to support custom labels, or we return a weird label because of some internal error), do we want to have to explicitly turn on support here too?

src/groundlight/client.py
```diff
@@ -72,6 +72,15 @@ def __init__(self, endpoint: Optional[str] = None, api_token: Optional[str] = No
...
         self.image_queries_api = ImageQueriesApi(self.api_client)
 
+    @classmethod
+    def _post_process_image_query(cls, iq: ImageQuery) -> ImageQuery:
+        """Post-process the image query so we don't use confusing internal labels.
+
+        TODO: Get rid of this once we clean up the mapping logic server-side.
+        """
+        iq.result.label = convert_internal_label_to_display(iq, iq.result.label)
+        return iq
```
mjvogelsong, line 82: Here are some alternatives to consider:
- we could put this conversion (and the `is_valid` validators) directly in the Pydantic `ImageQuery` model. The caveat is that `ImageQuery` and `ClassificationResult` are auto-generated from the OpenAPI spec, so we could accidentally lose the validator on next auto-generation.
- we could put this conversion (and the `is_valid` validators) in a light SDK Pydantic model wrapper or subclass. That way, we know it won't be overwritten by code geenration. But then we introduce another source of differences between the API and the SDK -- which can be good and bad.

src/groundlight/binary_labels.py
```diff
@@ -1,51 +1,98 @@
...
+
+
+DEPRECATED_LABEL_NAMES = {DeprecatedLabel.PASS, DeprecatedLabel.FAIL, DeprecatedLabel.NEEDS_REVIEW}
 
 
 def convert_internal_label_to_display(
     context: Union[ImageQuery, Detector, str],  # pylint: disable=unused-argument
     label: str,
 ) -> str:
+    """Convert a label that comes from our API into the label string that we show to the user.
```
mjvogelsong, line 37: I'm tempted to return enums instead of strings because our current service behavior does not support arbitrary strings, and it would steer users in the right direction. But I also understand that some users might not like enums, and prefer to work with strings.

src/groundlight/client.py
robotrapta: Either way it's a difference between the API & SDK, right?

I agree with not messing with the auto-generated code.  So putting it in a pydantic model seems reasonable.  But I'm also fine with this for now.

src/groundlight/binary_labels.py
```diff
@@ -1,51 +1,98 @@
...
+    if upper == Label.YES:
+        return DeprecatedLabel.PASS.value
+    if upper == Label.NO:
+        return DeprecatedLabel.FAIL.value
+    # TODO: Should we support the user adding an "UNSURE" label in the SDK?
+
+    raise ValueError(f'Invalid label string "{label}".  Must be one of {Label.YES.value},{Label.NO.value}.')
+
+
+def is_valid_display_result(result: Any) -> bool:
```
robotrapta, line 82: Is this method used outside tests?  If not, should it just go in the test files?

robotrapta (approved):
This all looks very reasonable to me.  Some responses to your thoughts within.

src/groundlight/binary_labels.py
```diff
@@ -1,51 +1,98 @@
...
-    if upper in {"FAIL", "NO"}:
-        return "FAIL"
-    raise ValueError(f'Invalid label string "{label}".  Must be one of YES,NO,PASS,FAIL')
+    if upper == Label.YES:
+        return DeprecatedLabel.PASS.value
+    if upper == Label.NO:
+        return DeprecatedLabel.FAIL.value
+    # TODO: Should we support the user adding an "UNSURE" label in the SDK?
+
+    raise ValueError(f'Invalid label string "{label}".  Must be one of {Label.YES.value},{Label.NO.value}.')
```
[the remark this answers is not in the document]
mjvogelsong: Yeah, I think `black` prefers this instead of escape characters.

PR #54, opened 2023-04-27
Faster async confident results

robotrapta:
This removes a delay before polling for confident results.  It was put in when the only source of async results was human, so it made sense to wait a while before polling.  Now ML results can be async, so this has been causing the SDK to return ML results much slower than necessary.


senorkumar (approved):
nice! I noticed this while building stuff, thanks for finding the root cause.

Issue #34, opened 2023-04-03
Switch from RGB to BGR color order to accept OpenCV images directly in `submit_image_query` 


robotrapta:
Right now we expect images to be in RGB color order if they're passed in as numpy arrays.  (This is what PIL / pillow use.)  But OpenCV uses BGR image format by default.

I suspect that if we get a numpy array, it is a lot more likely to have come from OpenCV than somebody converting a PIL image to a numpy array.  Implying we should expect BGR not RGB.

robotrapta:
Current consensus is that we should make this (breaking) change as part of the SDK v0.8.

brandon-groundlight:
Change made as of v0.8

PR #55, opened 2023-05-04
Numpy arrays must be in BGR order not RGB - Issue #34

brandon-groundlight:
Small fix to make image submission accept the bgr standard that opencv utilizes

### DIFF
--- a/docs/docs/installation/linux-windows-mac.md
+++ b/docs/docs/installation/linux-windows-mac.md
@@ -21,7 +21,7 @@ To install the Groundlight SDK using pip, run the following command in your term
 pip install groundlight
 ```
 
-If you're using `python3`, you might need to use `pip3` instead:
+If you're also using `python2` on your system, you might need to use `pip3` instead:
 
 ```bash
 pip3 install groundlight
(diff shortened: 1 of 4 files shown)


docs/docs/installation/linux-windows-mac.md
```diff
@@ -21,7 +21,7 @@ To install the Groundlight SDK using pip, run the following command in your term
 pip install groundlight
 ```
 
-If you're using `python3`, you might need to use `pip3` instead:
+If you're also using `python2` on your system, you might need to use `pip3` instead:
```
robotrapta, line 24: Good clarification

robotrapta (approved):
Awesome.  Thanks!

src/groundlight/images.py
```diff
@@ -49,7 +49,8 @@ def parse_supported_image_types(
         # Already in the right format
         return image
     if isinstance(image, np.ndarray):
-        return BytesIO(jpeg_from_numpy(image, jpeg_quality=jpeg_quality))
+        # Assume it is in BGR format from opencv
```
tyler-romero, line 52: Should we print a warning here the first time a user passes us a numpy array? Esp since this is a breaking change

PR #57, opened 2023-05-08
Docs - improved installation notes.  BGR vs RGB examples

robotrapta:
Examples of BGR / RGB split.  Better installation instructions for each platform, including arduino.


brandon-groundlight (approved):
The docs look much better! I'm not sure if you were intending to display the sections in a different order and intentionally chose to name "5-managing-confidence.md" with the leading number. But the compiled docs look good as they stand

PR #58, opened 2023-05-11
Replace all 'Test query?' questions in tests with 'Is there a dog?'

paulina-positronix:
Unfortunately something about this is failing

### DIFF
--- a/test/integration/test_groundlight.py
+++ b/test/integration/test_groundlight.py
@@ -6,12 +6,13 @@ from typing import Any
 
 import openapi_client
 import pytest
+from model import ClassificationResult, Detector, ImageQuery, PaginatedDetectorList, PaginatedImageQueryList
+
 from groundlight import Groundlight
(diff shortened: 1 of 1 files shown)


test/integration/test_groundlight.py
```diff
@@ -6,13 +6,15 @@
...
 import pytest
+
 from groundlight import Groundlight
 from groundlight.binary_labels import VALID_DISPLAY_LABELS, DeprecatedLabel, Label, convert_internal_label_to_display
 from groundlight.internalapi import NotFoundError
 from groundlight.optional_imports import *
 from groundlight.status_codes import is_user_error
 from model import ClassificationResult, Detector, ImageQuery, PaginatedDetectorList, PaginatedImageQueryList
 
+
```
mjvogelsong, line 17: I would just delete these added lines, that should fix the linter. Not sure why the auto-format bot isn't working.

mjvogelsong (approved):
Thanks for fixing this!

PR #59, opened 2023-05-19
Documenting how to use the environment variable to set the endpoint.

robotrapta:


docs/docs/building-applications/edge.md
```diff
@@ -1,10 +1,18 @@
...
 ```python
 from groundlight import Groundlight
 gl = Groundlight(endpoint="http://localhost:6717")
 ```
 
+or by setting the `GROUNDLIGHT_ENDPOINT` environment variable like:
+
+```bash
+export GROUNDLIGHT_ENDPOINT=http://localhost:6717
+python glapp.py
```
brandon-groundlight, line 14: took me a hot second to figure out glapp.py is just a generic name. Maybe your_app.py?

brandon-groundlight (approved):

docs/docs/building-applications/edge.md
robotrapta: Good idea.

PR #61, opened 2023-05-25
Improved debuggability

robotrapta:
The friendly string for the Groundlight SDK client object includes redacted API key and endpoint if configured.

Useful for figuring out why things aren't working.

### DIFF
--- a/src/groundlight/client.py
+++ b/src/groundlight/client.py
@@ -67,9 +67,17 @@ class Groundlight:
 
         configuration.api_key["ApiToken"] = api_token
 
+        self.api_token = api_token[:20] + "..."   # redact the secret
         self.api_client = GroundlightApiClient(configuration)
         self.detectors_api = DetectorsApi(self.api_client)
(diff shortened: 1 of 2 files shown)


src/groundlight/client.py
```diff
@@ -67,10 +67,18 @@ def __init__(self, endpoint: Optional[str] = None, api_token: Optional[str] = No
 
         configuration.api_key["ApiToken"] = api_token
 
+        self.api_token = api_token[:20] + "..."   # redact the secret
```
tyler-romero, line 70: Why `:20`? Is that partially revealing the secret? Or is the beginning of the key just boilerplate?

robotrapta: API tokens consist of a unique KSUID and a random secret.  The `:20` truncates to just the KSUID, which is what we store in the backend - so it's the recoverable / identifiable part of the token.

mjvogelsong:
Doing some old PR cleanup. Feel free to re-open.

PR #63, opened 2023-05-26
Allow specification of pipeline config when creating detector

tyler-romero:


src/groundlight/client.py
```diff
@@ -100,13 +100,13 @@ def create_detector(
         query: str,
         *,
         confidence_threshold: Optional[float] = None,
-        config_name: Optional[str] = None,
+        pipeline_config: Optional[str] = None,
```
tyler-romero, line 103: This is the only real change...

spec/public-api.yaml
```diff
@@ -265,13 +265,11 @@ components:
           default: 0.9
           description: If the detector's prediction is below this confidence threshold,
             send the image query for human review.
-        config_name:
+        pipeline_config:
```
tyler-romero, line 268: the main change of this PR. Question: should we just keep the field as "config_name". Also, are there any customers we know of who use the `config_name` field?

brandon-groundlight: None of our new customers have started using it as far as I know. Lkq would be the only one who could be using it

generated/docs/DetectorCreationInput.md
```diff
@@ -9,7 +9,7 @@ Name | Type | Description | Notes
 **query** | **str** | A question about the image. | 
 **group_name** | **str** | Which group should this detector be part of? | [optional] 
 **confidence_threshold** | **float** | If the detector&#39;s prediction is below this confidence threshold, send the image query for human review. | [optional]  if omitted the server will use the default value of 0.9
-**config_name** | **str, none_type** | (Advanced usage) If your account has multiple named ML configuration options enabled, you can use this field to specify which one you would like to use. | [optional] 
+**pipeline_config** | **str, none_type** | (Advanced usage) Configuration to instantiate a specific prediction pipeline. | [optional]
```
robotrapta, line 12: Should we be manually updating the docs in the `generated/docs` directory?

robotrapta (approved):
LGTM

tyler-romero:
The tests here are going to fail until corresponding updates have been made to the public API. Will land this after landing the changes to the public API.

generated/docs/DetectorCreationInput.md
tyler-romero: cc: @michael-groundlight who might know more about how these docs are used

mjvogelsong (approved):
Yep, looks good -- good to merge once the server-side fix is deployed.

PR #66, opened 2023-06-05
Answer vs Guess and UNCLEAR

robotrapta:
Proposal for v0.9 branch which separates the `answer` from the `guess` and includes `UNSURE` in addition to `YES` and `NO` as valid results.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ packages = [
     {include = "**/*.py", from = "src"},
 ]
 readme = "README.md"
-version = "0.8.0"
+version = "0.9.0"
 
 [tool.poetry.dependencies]
 certifi = "^2021.10.8"
--- a/src/groundlight/client.py
+++ b/src/groundlight/client.py
@@ -4,28 +4,22 @@ import time
 from io import BufferedReader, BytesIO
 from typing import Optional, Union
 
-from model import (
-    ClassificationLabel,
-    Detector,
-    ImageQuery,
-    PaginatedDetectorList,
-    PaginatedImageQueryList,
-    PartialImageQuery,
-)
+from model import (ClassificationLabel, Detector, ImageQuery,
+                   PaginatedDetectorList, PaginatedImageQueryList,
+                   PartialImageQuery)
 from openapi_client import Configuration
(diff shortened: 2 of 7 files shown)


pyproject.toml
```diff
@@ -9,7 +9,7 @@ packages = [
     {include = "**/*.py", from = "src"},
 ]
 readme = "README.md"
-version = "0.8.0"
+version = "0.9.0"
```
robotrapta, line 12: Arguably this isn't breaking because we're just adding new fields not removing the old ones.  But semantically it's a pretty big change.

src/groundlight/images.py
```diff
@@ -53,7 +53,7 @@ def parse_supported_image_types(
         return BytesIO(jpeg_from_numpy(image[:, :, ::-1], jpeg_quality=jpeg_quality))
     raise TypeError(
         (
-            "Unsupported type for image. Must be PIL, numpy (H,W,3) RGB, or a JPEG as a filename (str), bytes,"
-            " BytesIO, or BufferedReader."
+            f"Unsupported image type={type(image)}. Must be PIL, numpy (H,W,3) BGR, or JPEG as str (filename),"
```
robotrapta, line 56: This is a doc (exception msg) bug which should be back-ported to 0.8.

src/groundlight/client.py
```diff
@@ -206,26 +226,25 @@ def wait_for_confident_result(
...
-        delay = self.POLLING_INITIAL_DELAY
-        while time.time() < timeout_time:
-            current_confidence = image_query.result.confidence
-            if current_confidence is None:
-                logging.debug("Image query with None confidence implies human label (for now)")
+        start_time = time.time() 
+        next_delay = self.POLLING_INITIAL_DELAY
+        target_delay = 0
+        image_query = self._post_process_image_query(image_query)
+        while True:
```
robotrapta, line 233: I think there was a logic bug in the delay calculations in here which we should extract.

robotrapta:
Oh yeah, and the `List` method doesn't work yet.  I thought I'd get some discussion on the approach before ironing out all the details.

mjvogelsong:
Doing some old PR cleanup -- feel free to re-open.

PR #67, opened 2023-06-05
Fixing bug in timing of polling for confident result

robotrapta:
The logic for calculating delays in polling was wrong, so it was in a tight-loop to poll.  Now the logic is fixed, and also the debug logs are more clear about what's going on.

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -148,7 +148,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: get code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           ref: ${{ github.head_ref }}
       - name: install python
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -148,7 +148,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: get code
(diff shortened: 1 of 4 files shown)


robotrapta:
Converting to draft while tests are broken.

.github/workflows/cicd.yaml
```diff
@@ -148,7 +148,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: get code
-        uses: actions/checkout@v1
+        uses: actions/checkout@v3
```
robotrapta, line 151: This was causing the autoformat bot to fail to push.  Odd but true.

src/groundlight/client.py
```diff
@@ -206,26 +203,28 @@ def wait_for_confident_result(
...
+            sleep_time = max(target_delay - patience_so_far, 0)
             logger.debug(
-                (
-                    f"Polling for updated image_query because confidence {current_confidence:.3f} <"
-                    f" {confidence_threshold:.3f}"
-                ),
+                f"Polling ({target_delay:.1f}/{timeout_sec:.0f}s) {image_query} until"
+                f" confidence>={confidence_threshold:.3f}"
             )
-            time_left = max(0, time.time() - timeout_time)
```
robotrapta, line 225: Do you see the bug?

PR #68, opened 2023-06-07
Training test

brandon-groundlight:
Adds a test that models get trained after receiving 2 yes examples and 2 no examples


test/integration/test_groundlight.py
```diff
@@ -336,3 +336,44 @@ def test_submit_numpy_image(gl: Groundlight, detector: Detector):
     assert str(_image_query)
     assert isinstance(_image_query, ImageQuery)
     assert is_valid_display_result(_image_query.result)
+
+
+@pytest.mark.skipif(MISSING_NUMPY or MISSING_PIL, reason="Needs numpy and pillow")  # type: ignore
```
robotrapta, line 341: numpy isn't installed by default (because it's big).  So this test is going to get skipped a lot.  It might work properly in the places we need it to - we could check the CI/CD pipelines to make sure they're installing the extras.  But even if we do that, this test becomes somewhat fragile in that it needs `numpy` and can effectively "fail silently" if numpy isn't there.  Whereas `pillow` / PIL is installed by default.

So I'd really rather we get this functionality without relying on numpy.

test/integration/test_groundlight.py
```diff
@@ -336,3 +336,44 @@ def test_submit_numpy_image(gl: Groundlight, detector: Detector):
...
+def test_detector_improvement(gl: Groundlight, detector: Detector):
+    # test that we get confidence improvement after sending images in
+    # Pass two of each type of image in
+    import time
+
+    from PIL import Image
+
+    def submit_noisy_image(image):
```
robotrapta, line 349: So we want some way to randomly generate small image variations - numpy can add noise, and that's a good way.  But without numpy what can we do?  PIL/pillow has a number of methods which we can use as augmentations. How about some small random crop?  Or a small tweak to brightness?

test/integration/test_groundlight.py
```diff
@@ -336,3 +336,44 @@ def test_submit_numpy_image(gl: Groundlight, detector: Detector):
...
+        np_noise = np.random.normal(loc=0, scale=50, size=image.shape)
+        noisy_img = np.clip(image + np_noise, 0, 255).astype(np.uint8)
+        img_query = gl.submit_image_query(detector=detector.id, image=noisy_img)
```
robotrapta, line 352:
```suggestion
        img_query = gl.submit_image_query(detector=detector.id, image=noisy_img, wait=0)
```

Let's set the `wait` here because otherwise the system will wait for human review.

test/integration/test_groundlight.py
```diff
@@ -336,3 +336,44 @@ def test_submit_numpy_image(gl: Groundlight, detector: Detector):
...
+
+    dog = np.array(Image.open("test/assets/dog.jpeg"))[:, :, ::-1]
+    cat = np.array(Image.open("test/assets/cat.jpeg"))[:, :, ::-1]
+
+    dog_query_1 = submit_noisy_image(dog)
+    dog_query_2 = submit_noisy_image(dog)
+    cat_query_1 = submit_noisy_image(cat)
+    cat_query_2 = submit_noisy_image(cat)
+
+    gl.add_label(dog_query_1, "YES")
```
robotrapta, line 363: Let's merge these into `submit_noise_image` so the label is created immediately after the imagequery.  This will make the system more reliable -- because there's a much smaller gap of time when a cloud labeler might see this.

test/integration/test_groundlight.py
```diff
@@ -336,3 +336,44 @@ def test_submit_numpy_image(gl: Groundlight, detector: Detector):
...
+    gl.add_label(dog_query_2, "YES")
+    gl.add_label(cat_query_1, "NO")
+    gl.add_label(cat_query_2, "NO")
+
+    # wait to give enough time to train
+    wait_time = 300  # seconds
```
robotrapta, line 369: This is an awful long time to wait for the test to run.  It'll really slow down our pipelines.  BUT sometimes our pipelines do take several minutes to train an update.  They shouldn't - in fact, we should fail this test if it's not updated after I'd say 2 minutes.

Even still 2 minutes is a long time to wait.  So let's just wait 30 seconds, and then submit a couple imagequeries.  If they come back confident, success - we're done.  If they don't, then wait another 30 seconds and try again.  Repeat this a few times to wait up to 2 minutes.

test/integration/test_groundlight.py
```diff
@@ -336,3 +336,44 @@ def test_submit_numpy_image(gl: Groundlight, detector: Detector):
...
+    time.sleep(wait_time)
+
+    dog_query_3 = submit_noisy_image(dog)
+    cat_query_3 = submit_noisy_image(cat)
+    print(cat_query_1.result.confidence)
+    print(dog_query_1.result.confidence)
+    print(cat_query_3.result.confidence)
+    print(dog_query_3.result.confidence)
+    assert dog_query_3.result.confidence > dog_query_1.result.confidence
```
robotrapta, line 378: This criterion isn't strong enough IMHO.  Instead, let's set a fixed minimum expected confidence value like `0.6`.

robotrapta (changes requested):
A handful of suggestions within - I know this is probably pretty close to what we talked about, so thanks for that.  But looking through it I realize it should be more robust in a number of ways.

robotrapta (approved):
Great!

PR #70, opened 2023-06-08
Add Retries to the Python SDK with Exponential Backoff 

blaise-muhirwa:
This PR ensures that we retry to re-establish HTTP connections whenever we get status error codes in the range [500, 599]. We do that by exponentially backing off by a factor of 2 with a starting delay time of 100 milliseconds.


pyproject.toml
```diff
@@ -28,6 +28,7 @@ pytest = "^7.0.1"
 pytest-cov = "^3.0.0"
 pytest-markdown-docs = "^0.4.3"
 pytest-mock = "^3.10.0"
+requests-mock = "1.10.0"
```
robotrapta, line 31: Thanks for putting this in the `dev` dependencies.  We want to keep this library quite lightweight.  So be conscious of adding libraries to the main dependencies - especially big ones.

src/groundlight/internalapi.py
```diff
@@ -12,6 +14,7 @@
 from groundlight.status_codes import is_ok
 
 logger = logging.getLogger("groundlight.sdk")
+logger.setLevel(level=logging.DEBUG)
```
robotrapta, line 17: This should not get checked in.

src/groundlight/internalapi.py
```diff
@@ -70,6 +73,49 @@ class GroundlightApiClient(ApiClient):
     """
 
     REQUEST_ID_HEADER = "X-Request-Id"
+    MAX_CONNECTION_RETRY_ATTEMPTS = 4
+    # Initial delay time of 100 milliseconds
+    RETRY_INITIAL_DELAY = 0.1
+    RETRY_EXPONENTIAL_BACKOFF = 2
+    some_variable = 0
```
robotrapta, line 80: Work-in-progress code?  This should not get checked in.

src/groundlight/internalapi.py
```diff
@@ -70,6 +73,49 @@ class GroundlightApiClient(ApiClient):
...
+
+    # Defines the range of HTTP error status codes for which the API
+    # will attempt to re-establish a connection
+    STATUS_CODE_RANGE = range(500, 600)
```
robotrapta, line 84: Nice.

src/groundlight/internalapi.py
```diff
@@ -70,6 +73,49 @@ class GroundlightApiClient(ApiClient):
...
+
+    @staticmethod
+    def _retry_http_connection(callable):
```
robotrapta, line 87: Let's pull the decorator out into its own class.  That makes it easier to understand and test.

src/groundlight/internalapi.py
```diff
@@ -70,6 +73,49 @@ class GroundlightApiClient(ApiClient):
...
+            delay = self.RETRY_INITIAL_DELAY
+            retry_count = 0
+            while retry_count <= self.MAX_CONNECTION_RETRY_ATTEMPTS:
+                try:
+                    return callable(self, *args, **kwargs)
+
+                except requests.exceptions.HTTPError as error:
+                    status_code = error.response.status_code
+                    if status_code in self.STATUS_CODE_RANGE:
+                        logger.debug(
```
robotrapta, line 107: This should be at `warning` level.  It's fairly serious.

src/groundlight/internalapi.py
```diff
@@ -70,6 +73,49 @@ class GroundlightApiClient(ApiClient):
...
+                            f"Current HTTP response status: {status_code}."
+                            f"Remaining retries: {self.MAX_CONNECTION_RETRY_ATTEMPTS - retry_count + 1}"
+                        )
```
robotrapta, line 110: We should include `exc_info=True` in the logger call so the full stack trace and metdata are included in the logs.

src/groundlight/internalapi.py
```diff
@@ -70,6 +73,49 @@ class GroundlightApiClient(ApiClient):
...
+                        random_delay = random.uniform(0, delay)
+                        time.sleep(random_delay)
```
robotrapta, line 112: Perfect implementation.  I'd add a short comment saying this is implementing a "full-jitter" strategy.

src/groundlight/internalapi.py
```diff
@@ -70,6 +73,49 @@ class GroundlightApiClient(ApiClient):
...
+                retry_count += 1
+                delay *= self.RETRY_EXPONENTIAL_BACKOFF
+
+            raise InternalApiError("Maximum retries reached.")
```
robotrapta, line 116: When you re-raise you should do it in a way that wraps the original `requests` error.  I can't remember the right way to do this in py3.7 (lowest we support) but you either pass the `error` object into the constructor of the new exception, or there's some syntactic sugar like `InternalApiError(...) from error`.

src/groundlight/internalapi.py
```diff
@@ -122,19 +169,20 @@ def _add_label(self, image_query_id: str, label: str) -> dict:
...
         headers = self._headers()
-        response = requests.request("GET", url, headers=headers)
 
-        if not is_ok(response.status_code):
-            raise InternalApiError(
-                f"Error getting detector by name '{name}' (status={response.status_code}): {response.text}",
-            )
+        try:
+            response = requests.request("GET", url, headers=headers)
+            response.raise_for_status()
```
robotrapta, line 183: This is cleaner code, but is a functional change, and technically an incompatible change.  If somebody had code that caught `InternalApiError` it wouldn't work any more.

src/groundlight/internalapi.py
```diff
@@ -122,19 +169,20 @@ def _add_label(self, image_query_id: str, label: str) -> dict:
...
+        except requests.exceptions.HTTPError as error:
+            raise error
```
robotrapta, line 185: Catching an exception and immediately raising it doesn't do anything.  It would be reasonable to log a warning here.

test/unit/test_http_retries.py
```diff
@@ -0,0 +1,47 @@
...
+DEFAULT_CONFIDENCE_THRESHOLD = 0.9
+DETECTOR_NAME = "test detector"
+TOTAL_RETRIES = 4
+
+
+@pytest.fixture(name="gl")
+def groundlight_fixture() -> Groundlight:
+    "Creates a Groundlight client"
+    gl = Groundlight()
+    gl.DEFAULT_WAIT
```
robotrapta, line 15: What is this line supposed to do?

test/unit/test_http_retries.py
```diff
@@ -0,0 +1,47 @@
...
+def test_get_detector_by_name_attempts_retries(gl: Groundlight):
+    """Tests that call to `get_detector_by_name` retries to establish
+    a HTTP connection in case of a 500-599 response.
+    Here we are forcing the "execution" of requests.request to always 
+    return a status error code in range [500, 599]
+    """
+    status_codes = range(500, 600)
+    url = "https://api.groundlight.ai/device-api/v1/detectors?name=Test detector"
+
+    for status_code in status_codes:
```
robotrapta, line 36: I don't think we need to test all 100 return codes.  500 - 504 is fine, and 20x faster.  :)

robotrapta (changes requested):
Great start.  Retry logic is implemented perfectly!

A number of comments inline.  Let's move the decorator to its own class.  Make sure it covers all the API methods - not just a couple.  We should be able to do that in a general way without having to decorate lots of methods.

robotrapta (changes requested):
Great start.  Retry logic is implemented perfectly!

A number of comments inline.  Let's move the decorator to its own class.  Make sure it covers all the API methods - not just a couple.  We should be able to do that in a general way without having to decorate lots of methods.

test/unit/test_http_retries.py
```diff
@@ -0,0 +1,47 @@
...
+        with requests_mock.Mocker() as m:
+            m.get(url, status_code=status_code)
```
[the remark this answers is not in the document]
blaise-muhirwa: I ended up changing this to use the standard `unittest.mock`.

test/unit/test_http_retries.py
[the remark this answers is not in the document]
blaise-muhirwa: I added this as a TODO for now. I couldn't find a clean way to test the behavior where a request fails but subsequently succeeds.

src/groundlight/client.py
```diff
@@ -148,18 +157,21 @@ def get_or_create_detector(
...
 
+    @RequestsRetryDecorator()
     def list_image_queries(self, page: int = 1, page_size: int = 10) -> PaginatedImageQueryList:
         obj = self.image_queries_api.list_image_queries(page=page, page_size=page_size)
         image_queries = PaginatedImageQueryList.parse_obj(obj.to_dict())
         if image_queries.results is not None:
             image_queries.results = [self._fixup_image_query(iq) for iq in image_queries.results]
         return image_queries
 
+    @RequestsRetryDecorator()
```
robotrapta, line 174: This is too high a level for the retry decorator, because there's a bunch of logic going on inside here.  This method first calls the `POST` API method to create the imagequery, and then polls for a confident response.  If any of those return a 500, then the whole thing will retry, which is not what we want.  

Imagine if we get a 500 on one of the polling requests which is checking to see if a human label has showed up yet.  Having a retry at this level would result in a second imagequery being created.

src/groundlight/internalapi.py
```diff
@@ -67,9 +70,80 @@ def iq_is_confident(iq: ImageQuery, confidence_threshold: float) -> bool:
...
+        exponential_backoff: int = 2,
+        status_code_range: tuple = (500, 600),
+        max_retries: int = 3,
+    ):
+        self.initial_delay = initial_delay
+        self.exponential_backoff = exponential_backoff
+        self.status_code_range = range(*status_code_range)
+        self.max_retries = max_retries
+
+    def __call__(self, function: Callable[ParamsType, ReturnType]) -> Callable[ParamsType, ReturnType]:
```
robotrapta, line 111: The input/output types don't seem to be helpful here.  Can we just do this instead?

```suggestion
    def __call__(self, function: Callable) -> Callable:
```

src/groundlight/internalapi.py
```diff
@@ -67,9 +70,80 @@ def iq_is_confident(iq: ImageQuery, confidence_threshold: float) -> bool:
...
+    # pylint: disable=useless-super-delegation
+    def __init__(self, status=None, reason=None, http_resp=None):
+        super().__init__(status, reason, http_resp)
+
+
+ReturnType = TypeVar("ReturnType")
+ParamsType = ParamSpec("ParamsType")
+
+
+class RequestsRetryDecorator:
```
robotrapta, line 87: I like this.

robotrapta (changes requested):
The decorator class looks good.  The way it's applied definitely doesn't work everywhere - e.g. `submit_image_query` can get really confused like this.

Let's hop on a call and chat about putting it on `GroundlightApiClient.call_api`.

src/groundlight/internalapi.py
```diff
@@ -67,9 +71,76 @@ def iq_is_confident(iq: ImageQuery, confidence_threshold: float) -> bool:
...
+        @wraps(function)
+        def decorated(*args, **kwargs):
+            delay = self.initial_delay
+            retry_count = 0
+
+            while retry_count <= self.max_retries:
+                try:
+                    return function(*args, **kwargs)
+                except ApiException as e:
+                    is_target_error = (e.status is not None) and (e.status in self.status_code_range)
```
robotrapta, line 120:
```suggestion
                    is_retryable = (e.status is not None) and (e.status in self.status_code_range)
```
Slightly clearer name.

src/groundlight/internalapi.py
```diff
@@ -67,9 +71,76 @@ def iq_is_confident(iq: ImageQuery, confidence_threshold: float) -> bool:
     return iq.result.confidence >= confidence_threshold
 
 
-class InternalApiError(RuntimeError):
-    # TODO: We need a better exception hierarchy
-    pass
+class InternalApiError(ApiException, RuntimeError):
+    # TODO: We should really avoid this double inheritance since
```
robotrapta, line 75: I'd prefer to clean this up rather than make it uglier.  Changing the base class technically is a breaking change though.  So maybe we leave that for another PR.  This one is getting long in the tooth already.

src/groundlight/internalapi.py
```diff
@@ -79,7 +150,40 @@ class GroundlightApiClient(ApiClient):
...
+    # This represents the byte stream of the input image file.
+    # We cache the byte stream so that if we attempt to access
+    # the file again we can do so even when the file has already been closed.
+    CACHED_BYTE_STREAM = io.BytesIO()
+
+    @staticmethod
+    def get_file_data_and_close_file(file_instance: io.IOBase) -> bytes:
+        """Overrides the method from OpenAPI client in order to allow access to the
+        file byte stream even after the file has been closed.
+        NOTE: Assumes that file_instance is either io.BytesIO or io.BufferedReader
```
robotrapta, line 162: We should be explicit about this.  If it's not one of those two, we should raise an exception.

src/groundlight/internalapi.py
```diff
@@ -79,7 +150,40 @@ class GroundlightApiClient(ApiClient):
...
+        """
+        if file_instance.closed:
```
robotrapta, line 164: This logic isn't reliable - if the customer happens to pass in a file object which is already closed, then we'll do the wrong thing and re-send the previous image.

src/groundlight/internalapi.py
```diff
@@ -79,7 +150,40 @@ class GroundlightApiClient(ApiClient):
...
+            if isinstance(file_instance, BytesIO):
+                file_instance = BytesIO(GroundlightApiClient.CACHED_BYTE_STREAM.getvalue())
```
robotrapta, line 166: This is a clever way to work around the problem of trying to maintain state in a static method, but isn't thread safe.  And we already have use-cases deployed which are multi-threaded.  Which makes me think the strategy of overriding this `get_file_data_and_close_file ` method might not be the right way to go.

robotrapta (commented):
First off, good catch realizing that the `file_instance.read()` would fail on a retry because it got closed.  I think I probably would have missed that.  This is a clever solution, but I think it still doesn't work.  Primarily because it's not thread-safe, which is tricky because of the `staticmethod`.

Here's another idea of how to do it.  Since `IOBase` is this super generic base class / interface, and we know they're only calling `.read()` and `.close()` on it, let's just make our own subclass which is a thin wrapper around `bytes` and just returns the bytes on `read()` and does nothing on `close()`.  When we get the image, convert it into our own class, and then we don't need to worry about it getting closed because closing it does nothing.

src/groundlight/internalapi.py
blaise-muhirwa: I can submit a follow up PR sometime this week cleaning this up.

src/groundlight/internalapi.py
blaise-muhirwa: Good catch!

src/groundlight/images.py
```diff
@@ -1,10 +1,35 @@
...
+    maintain files in an open state. This is useful, in particular,
+    when we want to retry accessing the file without having to re-open it.
+    """
+
+    def __init__(self, data: Union[BufferedReader, BytesIO, bytes]) -> None:
+        super().__init__()
+        self._data = data
+
+    def read(self) -> bytes:
+        if isinstance(self._data, (BufferedReader, BytesIO)):
```
robotrapta, line 19: I don't think this works.  Because if you call `.read()` twice in a row on a BytesIO object, the second time, you get back `b''` - empty.  I think you should do this conversion (calling `read()`)in `__init__` to ensure you only call read() once.  And then read & getvalue can just return data.

src/groundlight/images.py
```diff
@@ -29,28 +54,29 @@ def jpeg_from_numpy(img: np.ndarray, jpeg_quality: int = 95) -> bytes:
...
         # Already in the right format
-        return image
+        return ByteStreamWrapper(data=image)
     if isinstance(image, np.ndarray):
         # Assume it is in BGR format from opencv
-        return BytesIO(jpeg_from_numpy(image[:, :, ::-1], jpeg_quality=jpeg_quality))
+        return ByteStreamWrapper(data=jpeg_from_numpy(image[:, :, ::-1], jpeg_quality=jpeg_quality))
     raise TypeError(
         (
             "Unsupported type for image. Must be PIL, numpy (H,W,3) RGB, or a JPEG as a filename (str), bytes,"
```
robotrapta, line 82: This should say BGR not RGB now right?

robotrapta (changes requested):
Pretty sure the wrapper won't work as needed.  

To check, can you write some tests that instantiate the wrapper object with either `bytes`, or a BytesIO or a ByteStream and confirm that you can call `read()` and `close()` on it repeatedly and that it'll give the right response?  (IIRC it's read & close, but if that's not right, make the test match what the generated API client is doing.)

src/groundlight/images.py
```diff
@@ -1,10 +1,33 @@
 import imghdr
-from io import BufferedReader, BytesIO
+from io import BufferedReader, BytesIO, IOBase
 from typing import Union
 
 from groundlight.optional_imports import Image, np
 
 
+class ByteStreamWrapper(IOBase):
```
robotrapta, line 8: Yeah!  This looks right.

src/groundlight/images.py
```diff
@@ -29,31 +52,32 @@ def jpeg_from_numpy(img: np.ndarray, jpeg_quality: int = 95) -> bytes:
 def parse_supported_image_types(
     image: Union[str, bytes, Image.Image, BytesIO, BufferedReader, np.ndarray],
     jpeg_quality: int = 95,
-) -> Union[BytesIO, BufferedReader]:
+) -> ByteStreamWrapper:
     """Parse the many supported image types into a bytes-stream objects.
     In some cases we have to JPEG compress.
```
robotrapta, line 57:
```suggestion
    In some cases we have to JPEG compress.
    We always return a ByteStreamWrapper, which materializes BytesIO objects into actual bytes arrays in memory.
    This is to simplify the retry logic (on server-side 500 errors) which gets messy if you have to re-open a BufferedReader or similar. 
```

src/groundlight/images.py
```diff
@@ -29,31 +52,32 @@ def jpeg_from_numpy(img: np.ndarray, jpeg_quality: int = 95) -> bytes:
...
-        return bytesio
+        return ByteStreamWrapper(data=bytesio)
     if isinstance(image, (BytesIO, BufferedReader)):
         # Already in the right format
-        return image
+        return ByteStreamWrapper(data=image)
     if isinstance(image, np.ndarray):
         # Assume it is in BGR format from opencv
-        return BytesIO(jpeg_from_numpy(image[:, :, ::-1], jpeg_quality=jpeg_quality))
+        return ByteStreamWrapper(data=jpeg_from_numpy(image[:, :, ::-1], jpeg_quality=jpeg_quality))
```
robotrapta, line 77:
```suggestion
        rgb_img = image[:, :, ::-1]  # swap color channels
        return ByteStreamWrapper(data=jpeg_from_numpy(rgb_img, jpeg_quality=jpeg_quality))
```
That one line is getting a bit much for me.

test/unit/test_imagefuncs.py
```diff
@@ -69,3 +70,32 @@ def test_pil_support_ref():
...
+    ByteStreamWrapper and get the same output.
+    """
+
+    def run_test(byte_stream: ByteStreamWrapper):
+        previous_bytes = byte_stream.read()
+
+        current_attempt, total_attempts = 0, 5
+
+        while current_attempt < total_attempts:
+            new_bytes = byte_stream.read()
```
robotrapta, line 87: Nice check.

robotrapta (approved):
Awesome!  I'm excited to use this.

PR #71, opened 2023-06-12
Update Poetry Version to Latest 

blaise-muhirwa:


.github/workflows/cicd.yaml
```diff
@@ -6,7 +6,6 @@ on:
 
 env:
   PYTHON_VERSION: "3.10"
-  POETRY_VERSION: "1.4.0"
```
mjvogelsong, line 9: I would recommend changing this to `POETRY_VERSION: "1.5.1"` (or whatever is latest), rather than removing the version.

Generally, it's easier to have consistent, reproducible environments when the dependency versions are explicit. For example, if poetry has a new release, we could have the case where we don't change anything, but something breaks -- that can be kinda confusing. By using an explicit version for poetry, we can reduce the chance of that.

There's always a drawback, though - the drawback is that you have to manually bump to upgrade. But in this case, it's a one-liner and I think it's worth it.

PR #73, opened 2023-06-20
Changes the default detector for tests to not raise queries to labelers

brandon-groundlight:
To my understanding, tests for our labellers should be elsewhere such as the canary tests. Most of the tests here that hit labellers are designed to test something else, and we shouldn't care if a human labeller sees them.


mjvogelsong (approved):
Yeah, good idea.

PR #74, opened 2023-06-27
Allow Not Submitting Image Queries for Human Review

blaise-muhirwa:


robotrapta (changes requested):
As discussed, let's send the query param to the server which tells the service not to request or allow human review.

PR #78, opened 2023-07-20
Fix Integration Test

blaise-muhirwa:
This just ensures that we account for the case when `ImageQuery.result.confidence = None`. Comparing None types with floating points has undefined behavior.


brandon-groundlight (approved):
Awesome, thanks for fixing this

PR #79, opened 2023-07-27
Add `patience_time` and `human_review` Query Parameters to the Image Queries Endpoint

blaise-muhirwa:
This is related to [Pk-2096](https://positronixcorp.atlassian.net/jira/software/projects/PK/boards/6/backlog?assignee=712020%3A7a2c1c3d-1f08-49ae-8c5d-28670834dd39&selectedIssue=PK-2096) and [PK-1855](https://positronixcorp.atlassian.net/jira/software/projects/PK/boards/6/backlog?assignee=712020%3A7a2c1c3d-1f08-49ae-8c5d-28670834dd39&selectedIssue=PK-1855). 

The PR uses the `wait` parameter from `submit_image_query` to set the `patience_time` in the backend API.


spec/public-api.yaml
```diff
@@ -132,6 +132,13 @@ paths:
           type: string
         description: Choose a detector by its ID.
         required: true
+      - in: query 
+        name: patience_time 
+        schema: 
+          type: number 
+          format: float 
+        description: How long to wait for a confident response. 
+        required: true
```
mjvogelsong, line 141: I'm thinking about whether we want this parameter to be required or not. My initial thought is to not make it required, so that client code written against the old API can still be valid. But there could be different interpretations of `patience_time=None` (either "return immediately", or "use the default of 30 seconds").

Anyway, API changes are important enough that we should get a second opinion from @tyler-romero / @robotrapta.

mjvogelsong (commented):
Yep, this is on the right track -- although I'm kinda thinking we should make the param not required.

Also, as you mention, we'll need to update janzu to accept the new param before we push this change.

mjvogelsong:
Camilo might be working on adding `patience_time` on the API side, so I suggest syncing with him for that.

spec/public-api.yaml
robotrapta: Agreed this should be an optional parameter here.  If not supplied to the SDK method, we shouldn't send it to the API.  So the API also needs to have it be optional.

robotrapta:
The reason `test-simple` is failing is because the backend doesn't have the updated parameters yet?

generated/openapi_client/api/image_queries_api.py
```diff
@@ -289,6 +297,8 @@ def submit_image_query(self, detector_id, **kwargs):
             detector_id (str): Choose a detector by its ID.
 
         Keyword Args:
+            human_review (bool): Allow image queries to be marked for no human review.. [optional]
```
robotrapta, line 300:
```suggestion
            human_review (bool): If set to `False` then unconfident ML predictions will not be escalated to human review.  [optional, defaults `True`]
```

mjvogelsong (approved):
This looks good to me -- we can merge after the backend parameters are deployed, and the tests pass.

PR #81, opened 2023-08-02
Inspections

timmarkhuff:
This PR adds several methods to Groundlight that are required for the RPC Server/URCap. They are intended to be released as undocumented features and only used to support the RPC Server/URCap on the SDK (for now at least).

Since the required functionality for these methods is not available in image_queries_api, they use the internal API directly.

New methods include:
start_inspection
update_inspection_metadata
stop_inspection
update_detector_confidence_threshold

There was also some logic added to submit_image_query related to inspections: if an image query is submitted with no inspection_id, the query will be submitted through image_queries_api (as it has been previously), however if submit_image_query is called with an inspection_id, we use the private API client instead.


src/groundlight/client.py
mjvogelsong, line 1: Can you add some tests that call the newly added methods and verify that the response is as expected?

src/groundlight/internalapi.py
```diff
@@ -225,3 +235,125 @@ def _get_detector_by_name(self, name: str) -> Detector:
...
+    def start_inspection(self) -> str:
+        """Starts an inspection, returns the ID."""
+        url = f"{self.configuration.host}/inspections"
+
+        headers = self._headers()
+
+        response = requests.request("POST", url, headers=headers, json={})
+
+        if not is_ok(response.status_code):
+            raise InspectionError(f"Error starting inspection. Status code: {response.status_code}")
```
mjvogelsong, line 271: Why not use the `InternalApiError` (or a subclass of it) for consistency?

src/groundlight/internalapi.py
```diff
@@ -225,3 +235,125 @@ def _get_detector_by_name(self, name: str) -> Detector:
...
+
+        return response.json()["id"]
+
+    @RequestsRetryDecorator()
+    def update_inspection_metadata(self, inspection_id: str, user_provided_key, user_provided_value) -> None:
```
mjvogelsong, line 276: I think the key should be a string.

```suggestion
    def update_inspection_metadata(self, inspection_id: str, user_provided_key: str, user_provided_value) -> None:
```

src/groundlight/internalapi.py
```diff
@@ -225,3 +235,125 @@ def _get_detector_by_name(self, name: str) -> Detector:
...
+        return response.json()["result"]
+
+    @RequestsRetryDecorator()
+    def update_detector_confidence_threshold(self, detector_id: str, confidence_threshold: float) -> None:
+        """Updates the confidence threshold of a detector."""
+
+        # The API does not validate the confidence threshold,
+        # so we will validate it here and raise an exception if necessary.
+        if confidence_threshold < 0 or confidence_threshold > 1:
+            raise ValueError(f"Confidence threshold must be between 0 and 1. Got {confidence_threshold}.")
```
mjvogelsong, line 348: Good idea!

src/groundlight/internalapi.py
```diff
@@ -225,3 +235,125 @@ def _get_detector_by_name(self, name: str) -> Detector:
...
+        if not is_ok(response.status_code):
+            raise InspectionError(
+                f"Error getting inspection details for inspection {inspection_id}. Status code: {response.status_code}"
+            )
+
+        payload = {}
+
+        # Set the user_provided_id_key and user_provided_id_value if they were not previously set.
+        response_json = response.json()
+        if not response_json["user_provided_id_key"]:
```
mjvogelsong, line 304: I'm not 100% sure -- will the `"user_provided_id_key"` key always be in the response (even if the value is `None`)?

To be safe, we could do `if not response_json.get("user_provided_id_key")`, which won't fail if the key doesn't exist.

mjvogelsong (changes requested):

mjvogelsong:
Sorry I clicked the button early!

My main request is to add some tests that use the new methods -- they should send (a) well-formed input and verify the response is as expected and (b) invalid input and make sure it raises the error as expected.

src/groundlight/client.py
```diff
@@ -249,3 +265,22 @@ def add_label(self, image_query: Union[ImageQuery, str], label: Union[Label, str
         api_label = convert_display_label_to_internal(image_query_id, label)
 
         return self.api_client._add_label(image_query_id, api_label)  # pylint: disable=protected-access
+
+    def start_inspection(self) -> str:
+        """Starts an inspection report and returns the id of the inspection."""
+        return self.api_client.start_inspection()
+
+    def update_inspection_metadata(self, inspection_id: str, user_provided_key, user_provided_value) -> None:
```
robotrapta, line 273: Let's add type-hints on `user_provided_*`.  I think they're both str, right?

src/groundlight/internalapi.py
```diff
@@ -20,6 +22,14 @@ class NotFoundError(Exception):
     pass
 
 
+class InspectionError(Exception):
```
robotrapta, line 25: When does this get raised?  Can we add a comment to that effect?  Also, it should probably inherit from something other than `Exception`, like `RuntimeError`.  Or better-yet one of our own API exception classes.

src/groundlight/internalapi.py
```diff
@@ -20,6 +22,14 @@ class NotFoundError(Exception):
...
+    pass
+
+
+class UpdateDetectorError(Exception):
```
robotrapta, line 29: Same comments as above.

robotrapta (commented):
Mostly done reading it, but here's comments so far.

src/groundlight/client.py
timmarkhuff: Yes, they are. I hesitated about this initially, because the code I was copying from the RPC server did not have type hints, but they will be treated as strings in the end, so I will add the type hints.

src/groundlight/internalapi.py
timmarkhuff: It should always be there. I copied this line of code from roobik and it worked there. But just to be safe, I will use .get

src/groundlight/internalapi.py
timmarkhuff: Yes, the are always strings. I will add the type hints.

src/groundlight/internalapi.py
timmarkhuff: Done. Using `InternalApiError`

src/groundlight/client.py
```diff
@@ -182,33 +183,48 @@ def submit_image_query(
           Any binary format must be JPEG-encoded already.  Any pixel format will get
           converted to JPEG at high quality before sending to service.
         :param wait: How long to wait (in seconds) for a confident answer.
+        :param inspection_id: The ID of the inspection (str) to associate with the image query.
```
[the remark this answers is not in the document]
timmarkhuff: Done

pyproject.toml
```diff
@@ -9,7 +9,7 @@ packages = [
     {include = "**/*.py", from = "src"},
 ]
 readme = "README.md"
-version = "0.10.1"
+version = "0.10.2"
```
robotrapta, line 12: I'd say we should call this `0.11` - not because it's backwards incompatible, but because if somebody asks "what version of the SDK do I need to use UR Cap?" it's a bit nicer to say "0.11" than "you need at least "0.10.2".

src/groundlight/client.py
timmarkhuff: This is done.

mjvogelsong (changes requested):
Looking good! Just a couple small changes requested, and then it should be good to go.

src/groundlight/client.py
```diff
@@ -222,7 +239,10 @@ def wait_for_confident_result(
         :param confidence_threshold: The minimum confidence level required to return before the timeout.
         :param timeout_sec: The maximum number of seconds to wait.
         """
-        # TODO: Add support for ImageQuery id instead of object.
+        # Convert from image_query_id to ImageQuery if needed.
+        if isinstance(image_query, str):
+            image_query = self.get_image_query(image_query)
```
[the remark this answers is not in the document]
timmarkhuff: Done

src/groundlight/internalapi.py
```diff
@@ -225,3 +227,170 @@ def _get_detector_by_name(self, name: str) -> Detector:
...
+        """Stops an inspection and raises an exception if the response from the server does not indicate success.
+        Returns a string that indicates the result (either PASS or FAIL). The URCap requires this.
+        """
+        url = f"{self.configuration.host}/inspections/{inspection_id}"
+
+        headers = self._headers()
+
+        # Closing an inspection generates a new inspection PDF. Therefore, if the inspection
+        # is already closed, just return "COMPLETE" to avoid unnecessarily generating a new PDF.
+        response = requests.request("GET", url, headers=headers)
```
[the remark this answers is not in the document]
timmarkhuff: done

src/groundlight/internalapi.py
```diff
@@ -225,3 +227,170 @@ def _get_detector_by_name(self, name: str) -> Detector:
...
+        )
+
+        # In the API, 'send_notification' is used to control human_review escalation. This will eventually
+        # be deprecated, but for now we need to support it in the following manner:
+        if human_review == "ALWAYS":
+            url += "&send_notification=True"
+        elif human_review == "NEVER":
+            url += "&send_notification=False"
+        else:
+            pass  # append nothing to the URL, allow "DEFAULT" behavior
```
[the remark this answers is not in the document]
timmarkhuff: done

src/groundlight/internalapi.py
```diff
@@ -225,3 +227,170 @@ def _get_detector_by_name(self, name: str) -> Detector:
...
+        headers["Content-Type"] = "image/jpeg"
+
+        response = requests.request("POST", url, headers=headers, data=body.read())
+
+        if not is_ok(response.status_code):
+            logger.info(response)
+            raise InternalApiError(
+                status=response.status_code,
+                reason=f"Error submitting image query with inspection ID {inspection_id} on detector {detector_id}",
+            )
```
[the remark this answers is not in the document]
timmarkhuff: done

src/groundlight/internalapi.py
```diff
@@ -225,3 +227,182 @@ def _get_detector_by_name(self, name: str) -> Detector:
...
+        else:
+            pass  # don't send the send_notifications param, allow "DEFAULT" behavior
+
+        headers = self._headers()
+        headers["Content-Type"] = "image/jpeg"
+
+        response = requests.request("POST", url, headers=headers, params=params, data=body.read())
+
+        if not is_ok(response.status_code):
+            logger.info(response)
```
mjvogelsong, line 267: I think you can delete this line, now that response is in the error.

mjvogelsong (approved):
Great! Thanks for the comprehensive tests, as well.

PR #83, opened 2023-08-09
Increase patience time in test to valid number

mjvogelsong:
We require patience_time to be greater than or equal to 10 on the server side, so we need to change a couple SDK tests.

Note: the "invalid: not an int" errors need to be fixed server-side.


robotrapta (approved):
LGTM

timmarkhuff:
Is this a duplicate of https://github.com/groundlight/python-sdk/pull/84?

PR #84, opened 2023-08-09
updated wait times on tests such that they can pass server size valid…

sunildkumar:
We are experiencing issues with our SDK tests as the `wait` parameter on `submit_image_query` is failing server side validation. For now, one must either not set `wait` or set `wait` to `None`, `0`, or an integer greater than or equal to `10`. 

If the `wait` is set to `0`, we will use the default `patience_time` of `30` seconds while returning synchronously in client code. (@blaise-muhirwa is working on this)

### DIFF
--- a/test/integration/test_groundlight.py
+++ b/test/integration/test_groundlight.py
@@ -6,16 +6,12 @@ from typing import Any
 
 import openapi_client
 import pytest
-from model import (ClassificationResult, Detector, ImageQuery,
-                   PaginatedDetectorList, PaginatedImageQueryList)
-
 from groundlight import Groundlight
-from groundlight.binary_labels import (VALID_DISPLAY_LABELS, DeprecatedLabel,
-                                       Label,
-                                       convert_internal_label_to_display)
+from groundlight.binary_labels import VALID_DISPLAY_LABELS, DeprecatedLabel, Label, convert_internal_label_to_display
 from groundlight.internalapi import NotFoundError
 from groundlight.optional_imports import *
 from groundlight.status_codes import is_user_error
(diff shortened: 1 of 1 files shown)


test/integration/test_groundlight.py
```diff
@@ -164,7 +164,7 @@ def test_get_detector_by_name(gl: Groundlight, detector: Detector):
 
 def test_submit_image_query_blocking(gl: Groundlight, detector: Detector):
     # Ask for a trivially small wait so it never has time to update, but uses the code path
-    _image_query = gl.submit_image_query(detector=detector.id, image="test/assets/dog.jpeg", wait=2)
+    _image_query = gl.submit_image_query(detector=detector.id, image="test/assets/dog.jpeg", wait=10)
```
timmarkhuff, line 167: The comment seems to be at odds with what the test does now (10 seconds isn't really trivially small). 

Do we still need this test? What does it accomplish exactly? I find it confusing, e.g.`assert str(_image_query)` Why do we check that the image query's string repr is not an empty string? The following line checks that the return value is an instance of an ImageQuery, which should be sufficient.

sunildkumar: Regarding the comment - I've update this to be more representative of the new test behavior

> Do we still need this test? What does it accomplish exactly?
I think this test is useful as it tests the bare minimum codepath we expect our customers to call, unlike future tests that are more complex. 

> Why do we check that the image query's string repr is not an empty string? ....
I think it is reasonable that a customer might attempt to cast our code's output to a string, so making sure that this cast is successful is reasonable.

PR #86, opened 2023-08-09
Switch Human Review Query Parameter from Boolean to String

blaise-muhirwa:
This PR changes the `human_review` query parameter such that it is passed in as a string instead of a boolean variable. The choices for `human_review` include `DEFAULT`, `ALWAYS`, and `NEVER`. 

It also ensures that the patience time that is sent to the server-side code is at least 10. 

Test cases are expected to pass once this change is reflected in the server code.


test/integration/test_groundlight.py
```diff
@@ -184,6 +184,15 @@ def test_submit_image_query_filename(gl: Groundlight, detector: Detector):
     assert is_valid_display_result(_image_query.result)
 
 
+def test_submit_image_query_with_no_human_review(gl: Groundlight, detector: Detector):
+    # For now, this just tests that the image query is created successfully. There should probably be a better
+    # way to check that the image query is never escalated for human review.
+    _image_query = gl.submit_image_query(detector=detector.id, image="test/assets/dog.jpeg", human_review="NEVER")
+    assert str(_image_query)
+    assert isinstance(_image_query, ImageQuery)
+    assert is_valid_display_result(_image_query.result)
```
mjvogelsong, line 193: We should test with each parameter 🙂

mjvogelsong, line 193: I think we can just test `assert is_valid_display_result(_image_query.result)` -- the other ones should only be tested in one place.

mjvogelsong (approved):
Yay, we have the green check!

PR #89, opened 2023-08-23
Command line client

brandon-groundlight:
Adds a command line tool named glcli that is available immediately after pip install. Only exposes basic commands but is pretty useful, say, if you have a detector and want to send an image directly to a detector. Can get a summary of the tool by running `glcli` in the terminal


robotrapta:
I love this idea!  I think I'd prefer to have the CLI tool named `groundlight` instead of `glcli`.  The latter is fairly clunky.  Nice thing about having a unique name is that we can use it with little fear of collision.

Also I have a vague memory of disliking `fire` over other similar options like `click` but I can't remember why right now.

pyproject.toml
```diff
@@ -20,6 +20,7 @@ python = ">=3.7.0,<4.0"
 python-dateutil = "^2.8.2"
 requests = "^2.28.2"
 urllib3 = "^1.26.9"
+fire = "^0.5.0"
```
mjvogelsong, line 23: I would recommend using [typer](https://typer.tiangolo.com/tutorial/) instead -- I'm not sure `fire` is actively maintained.

src/groundlight/client.py
```diff
@@ -303,3 +304,49 @@ def stop_inspection(self, inspection_id: str) -> str:
     def update_detector_confidence_threshold(self, detector_id: str, confidence_threshold: float) -> None:
         """Updates the confidence threshold of a detector given a detector_id."""
         self.api_client.update_detector_confidence_threshold(detector_id, confidence_threshold)
+
+
+def glcli():
```
mjvogelsong, line 309: I think it might be better to put this in a separate file, for separation of concerns.

mjvogelsong (changes requested):
I like the idea of a CLI! But I think we should use `typer` instead of `fire`.

src/groundlight/cli.py
```diff
@@ -0,0 +1,44 @@
...
+def class_func_to_cli(method):
+    """
+    Given the class method, simplify the typing on the inputs so that Typer can accept the method
+    """
+
+    @wraps(method)
+    def wrapper(*args, **kwargs):
+        print(method(*args, **kwargs))  # this is where we output to the console
+
+    # not recommended practice to directly change annotations, but gets around Typer not supporting Union types
```
mjvogelsong, line 21: Well that's annoying! Haha

src/groundlight/client.py
```diff
@@ -83,16 +83,20 @@ def _fixup_image_query(self, iq: ImageQuery) -> ImageQuery:  # pylint: disable=n
         return iq
 
     def get_detector(self, id: Union[str, Detector]) -> Detector:  # pylint: disable=redefined-builtin
+        """Get a detector by id"""
```
mjvogelsong, line 86: Side question - why is this a Union type to begin with? When would we want to get a detector by detector? Seems like someone copy-pasted that from submit_image_query or something like that.

src/groundlight/cli.py
```diff
@@ -0,0 +1,44 @@
+from functools import wraps
+from typing import Union
+
+import typer
+from typing_extensions import get_origin
+
+from groundlight import Groundlight
+
+cli_app = typer.Typer()
```
mjvogelsong, line 9: This change should make it so that `groundlight` or `groundlight -h` or `groundlight --help` all give the help message.

```suggestion
cli_app = typer.Typer(no_args_is_help=True, context_settings={"help_option_names": ["-h", "--help"]})
```

test/unit/test_cli.py
```diff
@@ -0,0 +1,86 @@
...
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        text=True,
+    )
+    assert completed_process.returncode == 0
+
+
+def test_bad_commands():
+    completed_process = subprocess.run(["groundlight"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
+    assert completed_process.returncode != 0
```
mjvogelsong, line 77: We want this (and `-h` and `--help`) to give the help message.

test/unit/test_cli.py
mjvogelsong, line 1: When testing, I did notice one annoying thing -- if I try to run `groundlight --help`, I get the "API token missing" error message. It would be nice if we could do the help command without needing an API token.

Also, if we could catch the exception and simply print out the API token message without the stack trace, I think that would be better UX.

Not sure if there's an easy way to do that -- if not, we can do that in a later PR.

```
Traceback (most recent call last):
  File "/Users/michael/workplace/python-sdk/src/groundlight/client.py", line 63, in __init__
    api_token = os.environ[API_TOKEN_VARIABLE_NAME]
  File "/Users/michael/.pyenv/versions/3.10.10/lib/python3.10/os.py", line 680, in __getitem__
    raise KeyError(key) from None
KeyError: 'GROUNDLIGHT_API_TOKEN'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/michael/workplace/python-sdk/src/groundlight/cli.py", line 33, in groundlight
    gl = Groundlight()
  File "/Users/michael/workplace/python-sdk/src/groundlight/client.py", line 65, in __init__
    raise ApiTokenError(
groundlight.client.ApiTokenError: No API token found. Please put your token in an environment variable named "GROUNDLIGHT_API_TOKEN". If you don't have a token, you can create one at https://app.groundlight.ai/reef/my-account/api-tokens
```

mjvogelsong (approved):
I think this is a great idea! I had a few comments on UX.

src/groundlight/client.py
brandon-groundlight: I would think so, except there's the explicit short circuit logic below so this had a bit more thought than just a copy paste

test/unit/test_cli.py
brandon-groundlight: It doesn't feel complicated, but after playing with it for a bit it turns out to be trickier than I first thought. I'll make another PR for this

PR #90, opened 2023-08-24
Update certifi library to newer version

robotrapta:
We were depending on an older version of `certifi` which contains root certificates.  This updates the dependency to a newer version without known security isues.

Perhaps more importantly, because we were using semver's `^` we couldn't actually update to the newest version.  Now we use `>=` to allow it to use each year's newest versions going forwards.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,6 +12,7 @@ readme = "README.md"
 version = "0.11.0"
 
 [tool.poetry.dependencies]
+# For certifi, use ">=" instead of "^" since it upgrades its "major version" every year
 certifi = ">=2022.12.07"
 frozendict = "^2.3.2"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -12,7 +12,8 @@ readme = "README.md"
 version = "0.11.0"
 
 [tool.poetry.dependencies]
-certifi = "^2021.10.8"
+# For certifi, use ">=" instead of "^" since it upgrades its "major version" every year
+certifi = ">=2022.12.07"
```
mjvogelsong, line 16: I think we want `>=2023.7.22`: https://github.com/groundlight/python-sdk/security/dependabot/12

robotrapta: Thanks.  I'll update.

mjvogelsong (approved):
LGTM!

PR #95, opened 2023-09-09
Add support for png images

brandon-groundlight:
We won't make use of the alpha channel, but we can truncate it and accept the RGB data. This will work for images that don't have completely transparent parts, possibly reducing friction for first time users


test/integration/test_groundlight.py
mjvogelsong, line 1: I would also add a test specifically for `bytestream_from_filename()` in addition to the API-level test. It should verify the color channels are as expected after conversion.

mjvogelsong (commented):
Great, this will be more convenient for users. I do think we should add another test, though.

test/unit/test_imagefuncs.py
```diff
@@ -26,6 +26,22 @@ def test_jpeg_from_numpy():
...
+    images_streams.append(bytestream_from_filename("test/assets/cat.png", jpeg_quality=95))
+    for i in images_streams:
+        assert isinstance(i, ByteStreamWrapper)
+        image = Image.open(i)
+        assert image.mode == "RGB"
+
+    # pixel based test, verified the image is correct by eye, then got a pixel whose value to check against
+    png_bytestream = bytestream_from_filename("test/assets/cat.png", jpeg_quality=95)
+    png_image = Image.open(png_bytestream)
+    assert png_image.getpixel((200, 200)) == (215, 209, 197)
```
mjvogelsong, line 42: You might want to get a pixel that has different height and width values, just so that we don't accidentally miss an issue of flipping dimensions.

mjvogelsong (approved):
Thanks! One small adjustment to the test and then it should be good!

PR #97, opened 2023-09-15
Patience time

brandon-groundlight:
A pass at implementing wait and patience time in the SDK according to the [design doc](https://docs.google.com/document/d/1I1Vwnt-wz7z_oTuKMfNid4YyMj0KBvGDrevnisgB6pE/edit).


brandon-groundlight:
This isn't quite what we laid out in the planning document, I was struggling a bit because it seemed like ask_ml and ask_confident would devolve into each other, each being a limiting case of the other. 

In this version, the goal is to make the three different ask_* functions friendly user entry points. submit_image_query remains the workhorse submission function with the full feature set, but we can encourage use of the ask_* functions, none of which require knowledge of both wait_time and patience_time.

mjvogelsong:
I think https://github.com/groundlight/python-sdk/pull/99 is taking the place of this PR, so I'll close this one. Feel free to re-open if I'm mistaken!

PR #99, opened 2023-10-05
Add ask_confident and ask_ml

brandon-groundlight:
Adding ask_confident and ask_fast as two simplified versions of submit_image query.
In the process, I separate patience time from wait time and add confidence level to submit image query


src/groundlight/client.py
```diff
@@ -43,6 +43,7 @@ class Groundlight:
     """
 
     DEFAULT_WAIT: float = 30.0
+    DEFAULT_PATIENCE: float = 30.0
```
robotrapta, line 46: At first blush I'd think we shouldn't need this in the SDK.  Detectors have a patience time configured in the cloud.  If customers don't specify patience here, it should always use whatever's in the cloud.

src/groundlight/client.py
```diff
@@ -172,11 +173,62 @@ def list_image_queries(self, page: int = 1, page_size: int = 10) -> PaginatedIma
             image_queries.results = [self._fixup_image_query(iq) for iq in image_queries.results]
         return image_queries
 
+    def ask_confident(
+        self,
+        detector: Union[Detector, str],
+        image: Union[str, bytes, Image.Image, BytesIO, BufferedReader, np.ndarray],
+        wait: Optional[float] = None,
+    ) -> ImageQuery:
```
robotrapta, line 181: Intuitively I feel like I should be able to specify the confidence threshold here.  I know it's set on the detector.  But IIRC the API lets us override it for a specific IQ.  And given the method signature it feels like I should be able to include it.

src/groundlight/client.py
```diff
@@ -172,11 +173,62 @@ def list_image_queries(self, page: int = 1, page_size: int = 10) -> PaginatedIma
...
+          converted to JPEG at high quality before sending to service.
+        :param wait: How long to wait (in seconds) for a confident answer.
+        """
+        return self.submit_image_query(
+            detector,
+            image,
+            wait=wait,
+        )
+
+    def ask_fast(
```
robotrapta, line 200: I think I still prefer `ask_ml` here.

src/groundlight/client.py
```diff
@@ -172,11 +173,62 @@ def list_image_queries(self, page: int = 1, page_size: int = 10) -> PaginatedIma
...
+        self,
+        detector: Union[Detector, str],
+        image: Union[str, bytes, Image.Image, BytesIO, BufferedReader, np.ndarray],
+        wait: Optional[float] = None,
```
robotrapta, line 204: Semantically, what is "wait" supposed to do here?  We're asking it for the fastest possible answer, so what is this doing?

src/groundlight/client.py
```diff
@@ -191,6 +243,8 @@ def submit_image_query(  # noqa: PLR0913 # pylint: disable=too-many-arguments
           Any binary format must be JPEG-encoded already.  Any pixel format will get
           converted to JPEG at high quality before sending to service.
         :param wait: How long to wait (in seconds) for a confident answer.
+        :param patience_time: How long Groundlight should work to generate a confident answer, even working beyond when wait time
```
robotrapta, line 246:
```suggestion
        :param patience_time: How long Groundlight should work to generate a confident answer, even working beyond the specified wait time
```

src/groundlight/client.py
```diff
@@ -172,11 +173,62 @@ def list_image_queries(self, page: int = 1, page_size: int = 10) -> PaginatedIma
...
+          - PIL Image
+          Any binary format must be JPEG-encoded already.  Any pixel format will get
+          converted to JPEG at high quality before sending to service.
+        :param wait: How long to wait (in seconds) for a confident answer.
+        """
+        return self.submit_image_query(
+            detector,
+            image,
+            wait=wait,
+            confidence_threshold=0.5,
```
robotrapta, line 222: I don't think this is right.  Because this updates the permanent threshold for this IQ, which will cause it to never get escalated.  I think the way to do this is to set `wait=0` which causes `submit_image_query` to never go into polling mode.

The exception would be if the initial response doesn't have a real prediction yet (0.5) because the ML took too long, then this method should poll until we get something back.  I suppose we could have a `wait` timeout there.

src/groundlight/client.py
```diff
@@ -200,16 +254,15 @@ def submit_image_query(  # noqa: PLR0913 # pylint: disable=too-many-arguments
...
         detector_id = detector.id if isinstance(detector, Detector) else detector
 
         image_bytesio: ByteStreamWrapper = parse_supported_image_types(image)
 
         params = {"detector_id": detector_id, "body": image_bytesio}
-        if wait == 0:
-            params["patience_time"] = self.DEFAULT_WAIT
-        else:
-            params["patience_time"] = wait
+        params["patience_time"] = patience_time
```
robotrapta, line 265: I think we'd be better off leaving this blank if it wasn't set in python.  This will have the effect of overriding the server-configured patience time to 30, even if the customer has customized it on the website, if they don't include patience when calling the SDK.

robotrapta (changes requested):
A couple things I think should work differently.

src/groundlight/client.py
brandon-groundlight: The logic fundamentally looks for the first answer we get back, which is why I like ask_fast. 99.9% of the time it is de facto asking ML because the ML returns faster than anything else. I only worry that this could set up some confusion down the road if we have to explain to people why the first result is (almost) the same as the ML result.

I won't dig in too much on this though, it's updated to ask_ml in the latest version now.

src/groundlight/client.py
brandon-groundlight: Only for the edge case you mention in the other comment. Would changing "wait" to "max_wait" everywhere be clearer?

src/groundlight/client.py
```diff
@@ -191,6 +250,8 @@ def submit_image_query(  # noqa: PLR0913 # pylint: disable=too-many-arguments
           Any binary format must be JPEG-encoded already.  Any pixel format will get
           converted to JPEG at high quality before sending to service.
         :param wait: How long to wait (in seconds) for a confident answer.
+        :param patience_time: How long Groundlight should work to generate a confident answer, even working beyond the specified wait time
```
robotrapta, line 253: Should mention that this is optional, and that the detector’s default will be used if this is not specified.

src/groundlight/client.py
```diff
@@ -191,6 +250,8 @@ def submit_image_query(  # noqa: PLR0913 # pylint: disable=too-many-arguments
...
+        :param confidence_threshold: If set, override the detector's confidence threshold for this query.
```
robotrapta, line 254: I like this wording - let’s use this for “patience_time” too.  “If set, overrides…”

src/groundlight/client.py
```diff
@@ -226,11 +285,14 @@ def submit_image_query(  # noqa: PLR0913 # pylint: disable=too-many-arguments
...
-            image_query = self.wait_for_confident_result(image_query, confidence_threshold=threshold, timeout_sec=wait)
+            if confidence_threshold is None:
+                threshold = self.get_detector(detector).confidence_threshold
+            else:
+                threshold = confidence_threshold
+            image_query = self._wait_for_confident_result(image_query, confidence_threshold=threshold, timeout_sec=wait)
         return self._fixup_image_query(image_query)
 
-    def wait_for_confident_result(
+    def _wait_for_confident_result(
```
robotrapta, line 295: Should we hide this method?  I think it’s quite reasonable that customers will want to use this directly.  We have a bunch ourselves, but a lot of that was because the `ask_` methods weren’t available.  But I think, e.g. in the ask_async example with two threads - one thread just submits with ask_async, and the other one uses wait_for_confident to analyze.

src/groundlight/client.py
```diff
@@ -270,6 +332,37 @@ def wait_for_confident_result(
             image_query = self._fixup_image_query(image_query)
         return image_query
 
+    def _wait_for_ml_result(self, image_query: Union[ImageQuery, str], timeout_sec: float = 30.0) -> ImageQuery:
```
robotrapta, line 335: Can we merge this with “wait_for_confident”?  I imagine there’s a lot of duplicated logic in terms of retry and back off.

robotrapta (approved):
This looks good!  My biggest requests are to leave `wait_for_confident` as not hidden. (Changing it will break some internal apps, so we’d definitely need to bump to 0.12 for this.). And if it’s not too hard to merge wait_for_ml into it to reduce duplicated polling logic.

src/groundlight/client.py
brandon-groundlight: I merged them in an earlier draft but actually opted for duplicity in favor of better logging. I'm going back to the previous version with them merged but would love any tips on how to provide the same level of specificity while logging if the combined logic is indifferent to being called from wait_for_ml or from wait_for_confident

src/groundlight/client.py
brandon-groundlight: It's more of a gut instinct that made me want to hide it originally. I suspect that any use case someone has for it we'll want to build syntactic sugar around it like we're doing for the ask methods. The ask_async method may be the exception, so I'll make this unhidden for now.

PR #100, opened 2023-10-06
Add support for sphinx docs and set up simple GHA 

blaise-muhirwa:
This PR adds support for Sphinx docs. Sphinx supports two styles for generating docs: 

* reStructuredText (reST) Format
* Google-style doc strings

I've re-structured comments in `groundlight/client.py` to follow the reST format. Both formats are pretty strict. For instance, you have to separate function parameter descriptions by an empty space, no mixing up of tabs and space, etc. Also example code has to follow a strict format, but at the end of the day, it's worth it since the docs look nice. 

To generate the docs, you can run 

```
# This will create a directory called build
$ poetry run make html
$ open build/html/index.html
```

This will render the docs in your browser. 

NOTE: This PR does not link the docs to our actual website. That will be the next step to take from here. I've also set up a simple GitHub action to test that docs build properly. Moving forward I think we should make this action publish the docs every time we do an SDK release.


pyproject.toml
```diff
@@ -17,11 +17,13 @@ certifi = ">=2023.7.22"
 frozendict = "^2.3.2"
 pillow = "^9.0.0" # TODO: We may want to mark pillow (and numpy) as extra (https://python-poetry.org/docs/master/pyproject#extras)
 pydantic = "^1.7.4"
-python = ">=3.7.0,<4.0"
+python = ">=3.9.0,<4.0"
 python-dateutil = "^2.8.2"
 requests = "^2.28.2"
 urllib3 = "^1.26.9"
 typer = "^0.9.0"
+Sphinx = "7.2.6"
```
robotrapta, line 25: These should be moved into "dev" dependencies - we don't need them in the SDK at runtime, which is what happens in this section.

pyproject.toml
```diff
@@ -17,11 +17,13 @@ certifi = ">=2023.7.22"
 frozendict = "^2.3.2"
 pillow = "^9.0.0" # TODO: We may want to mark pillow (and numpy) as extra (https://python-poetry.org/docs/master/pyproject#extras)
 pydantic = "^1.7.4"
-python = ">=3.7.0,<4.0"
+python = ">=3.9.0,<4.0"
```
blaise-muhirwa, line 20: sphinx requires at least python 3.9 that's why.

sphinx_docs/conf.py
```diff
@@ -0,0 +1,37 @@
...
+
+import sys
+import os
+
+sys.path.insert(0, os.path.abspath("../src"))
+
+project = "Groundlight Python SDK"
+copyright = "2023, Groundlight AI <support@groundlight.ai>"
+author = "Groundlight AI <support@groundlight.ai>"
+release = "0.11.1"
```
robotrapta, line 18: We need to make sure this gets automatically updated with the current release version.  There's some trick which sets `groundlight.__version__` to the release number - we should use the same kind of thing here.  Otherwise this will get out of sync.

sphinx_docs/conf.py
```diff
@@ -0,0 +1,37 @@
...
+
+# -- General configuration ---------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#general-configuration
+
+# sphinx.ext.autodoc is a Sphinx extension that automatically documents Python modules.
+# sphinx.ext.napoleon is a Sphinx extension that enables Sphinx to understand Google style docstrings.
+# We are using the `reStructuredText` format for docstrings instead of google style, but functions that
+# use google style docstrings will still be documented correctly.
+extensions = ["sphinx.ext.autodoc", "sphinx.ext.napoleon"]
```
robotrapta, line 27: I don't think we're using Google style docstrings are we?  I think the `:param foo:` is reStructuredText

sphinx_docs/index.rst
```diff
@@ -0,0 +1,29 @@
+.. Groundlight Python SDK documentation master file, created by
+   sphinx-quickstart on Fri Oct  6 13:47:27 2023.
+   You can adapt this file completely to your liking, but it should at least
+   contain the root `toctree` directive.
```
robotrapta, line 4: Any reason to keep this?

src/groundlight/client.py
```diff
@@ -77,13 +100,23 @@ def __init__(self, endpoint: Optional[str] = None, api_token: Optional[str] = No
...
 
     def get_detector(self, id: Union[str, Detector]) -> Detector:  # pylint: disable=redefined-builtin
-        """Get a detector by id"""
+        """
+        Get a detector by id
+
+        :param id: the detector id
+        :type id: str or Detector
+
+        :return Detector
```
robotrapta, line 117: This implies we should include sphinx docs for the `Detector` class too.

src/groundlight/client.py
```diff
@@ -143,29 +228,44 @@ def get_or_create_detector(
...
 
     def get_image_query(self, id: str) -> ImageQuery:  # pylint: disable=redefined-builtin
-        """Get an image query by id"""
+        """
+        Get an image query by id
+
+        :param id: the image query id
+        :type id: str
+
+        :return ImageQuery
```
robotrapta, line 249: And `ImageQuery` class as well I think.

.github/workflows/sphinx_docs.yaml
robotrapta, line 1: We need to figure out how to merge this into the existing docs build for `code.groundlight.ai`.

robotrapta (commented):
Exciting start!  Some comments below.

.github/workflows/sphinx_docs.yaml
robotrapta: We can do this as a follow-up PR though.

sphinx_docs/index.rst
blaise-muhirwa: None! Getting rid of it.

sphinx_docs/conf.py
blaise-muhirwa: We're not. I had left it there to show how we would support it if we so desired.

pyproject.toml
```diff
@@ -16,20 +16,24 @@ version = "0.11.1"
...
+pydantic = ">=2.0,<3.0.0"
+python = ">=3.9.0,<4.0"
 python-dateutil = "^2.8.2"
 requests = "^2.28.2"
 urllib3 = "^1.26.9"
 typer = "^0.9.0"
 
 [tool.poetry.group.dev.dependencies]
-datamodel-code-generator = "^0.12.1"
+datamodel-code-generator = "^0.22.1"
```
blaise-muhirwa, line 27: Had to increase version in order to get this to play nice with other dependencies that require certain versions of pydantic.

.github/workflows/sphinx_docs.yaml
```diff
@@ -0,0 +1,37 @@
+name: Build Documentation
```
robotrapta, line 1:
```suggestion
name: Build API Reference docs using sphinx 
```

.github/workflows/sphinx_docs.yaml
```diff
@@ -0,0 +1,37 @@
...
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main 
+
+env:
+  PYTHON_VERSION: "3.10"
+  POETRY_VERSION: "1.5.1"
```
robotrapta, line 13: 1.6.1 is newer right?

.github/workflows/sphinx_docs.yaml
```diff
@@ -0,0 +1,37 @@
...
+          python-version: ${{ env.PYTHON_VERSION }}
+      - name: install poetry 
+        uses: snok/install-poetry@v1
+        with:
+          version: ${{ env.POETRY_VERSION }}
+
+      - name: Install dependencies
+        # This should really just install sphinx. I'll have to add a separate group 
+        # for sphinx dependencies, so we don't have to install everything. 
+        run: make install-dev
```
mjvogelsong, line 33: Yeah, I think something like `poetry add --group docs sphinx` should work.

Makefile
```diff
@@ -47,3 +50,25 @@ lint: install-lint  ## Run linter to check formatting and style
...
+SOURCEDIR     = sphinx_docs
+BUILDDIR      = build
+
+# Put it first so that "make" without argument is like "make help".
+help:
+	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+
+.PHONY: help Makefile
+
+# Catch-all target: route all unknown targets to Sphinx using the new
```
robotrapta, line 71: Does this mean if you type `make foobartypo` for the SDK it launches sphinx for something?  I'm not sure why we'd want that - seems like it would be more confusing than helpful.

.github/workflows/sphinx_docs.yaml
```diff
@@ -0,0 +1,37 @@
...
+      
+      - name: Build documentation
+        run: |
+          poetry run make html
```
mjvogelsong, line 37: I recommend creating a `make apidocs` target in the `Makefile`.

pyproject.toml
mjvogelsong: Some customers might still be using python 3.8 (it's not end-of-life yet - https://endoflife.date/python). So I'd recommend seeing if you can use a looser lower limit on the sphinx dependency that allows us to use `python >=3.8.0,<4.0`

robotrapta: This is not good.  Canceling support for py3.7 and py3.8 is a big deal.  py3.7 is pretty old, but py3.8 is still default on some popular systems.

We should figure out how to build the docs on a newer python without cutting off SDK support.

pyproject.toml
mjvogelsong, line 1: Can you run `make format`? It should sort this toml file.

robotrapta (changes requested):
We can't cut off py3.7 and py3.8 support.

mjvogelsong (changes requested):
I'm really excited about having sphinx docs, it's a great step!

My main request is to find a way to still allow python 3.8+.

Makefile
blaise-muhirwa: This was autogenerated, but I also think it's more confusing than helpful, so removing it!

src/groundlight/client.py
```diff
@@ -76,14 +101,25 @@ def __init__(self, endpoint: Optional[str] = None, api_token: Optional[str] = No
         self.detectors_api = DetectorsApi(self.api_client)
         self.image_queries_api = ImageQueriesApi(self.api_client)
 
-    def _fixup_image_query(self, iq: ImageQuery) -> ImageQuery:  # pylint: disable=no-self-use
-        """Process the wire-format image query to make it more usable."""
+    @staticmethod
+    def _fixup_image_query(iq: ImageQuery) -> ImageQuery:
```
blaise-muhirwa, line 105: Made this static to make pylint happy.

.github/workflows/sphinx_docs.yaml
```diff
@@ -0,0 +1,35 @@
...
+        uses: snok/install-poetry@v1
+        with:
+          version: ${{ env.POETRY_VERSION }}
+
+      - name: Install dependencies 
+        run: make install-sphinx-deps
+      
+      - name: Build API documentation
+        run: |
+          poetry run make apidocs
```
mjvogelsong, line 35: Can you move the `poetry run` inside the make target?

```suggestion
          make apidocs
```

pyproject.toml
```diff
@@ -39,6 +39,13 @@ ruff = "^0.0.261"
 toml-sort = "^0.23.0"
 types-requests = "^2.28.11.17"
 
+[tool.poetry.group.sphinx-deps.dependencies]
+python = ">=3.9.0,<4.0"
```
mjvogelsong, line 43: I think this works, but just double-check that we can still install the main package with python 3.8. The easiest way might be to temporarily comment out this line in the github action (https://github.com/groundlight/python-sdk/blob/main/.github/workflows/cicd.yaml#L89), and see it run with python 3.7-3.11.

mjvogelsong (approved):
Great! Just one thing to double-check, and then it's good to merge.

pyproject.toml
blaise-muhirwa: Verified!

.github/workflows/sphinx_docs.yaml
```diff
@@ -0,0 +1,35 @@
+name: Build API Reference docs using sphinx 
+
+on:
+  push:
+    branches:
+      - main
```
robotrapta, line 6: It would make sense to have this workflow run on PR as well - as a test that sphinx is working and happy.

robotrapta (approved):
Great!

PR #102, opened 2023-10-09
ask_async

sunildkumar:
This enables customers to query `groundlight` asynchronously. It adds `want_async` as a new parameter on `submit_image_query` and  implements `ask_async`, a convenient interface to submit queries asynchronously.

Blockers to merging this:
~~1. Merge my corresponding server-side PR the enables async IQ submissions.~~
~~2. Make changes based on updates to the BE~~
3. Fix PR comments 
4. Show that new SDK tests pass after merge, right now they fail because the server side code to support this feature isn't deployed. 
5. How do we want to document this added functionality? I will have to add to the docs reference in sphinx once @blaise-muhirwa sets that up :). I should definitely show a small example here about how you need to query for results after querying async.
6. Do we want to add a section to `docs/building-applications` with a small demo of how one can use `ask_async`?


src/groundlight/client.py
sunildkumar, line 1: In addition to this code, let me know what you think about this following:

1. How do we want to document this added functionality? I will have to add to the docs reference in sphinx once Blaise merges this. I should definitely show a small example here about how you need to query for results after querying async. 
2. Do we want to add a section to docs/building-applications with a small demo of how one can use ask_async?

spec/public-api.yaml
```diff
@@ -1,3 +1,4 @@
+# TODO - this is super out of date. Is it more misleading than helpful?
```
robotrapta, line 1: Good question.

src/groundlight/client.py
robotrapta: I think we should definitely sprinkle it into the top-level docs.  I’d describe an application where you have one thread uploading images and another thread processing the results.

src/groundlight/client.py
```diff
@@ -76,10 +71,19 @@ def __init__(self, endpoint: Optional[str] = None, api_token: Optional[str] = No
...
-        """Process the wire-format image query to make it more usable."""
+    def _fixup_image_query(self, iq: ImageQuery, want_async: bool = False) -> ImageQuery:  # pylint: disable=no-self-use
+        """
+        Post process an image query prior to returning it to the user.
+        :param iq: The image query to fix up.
+        :param want_async: post processing for want_async=True option in submit_image_query
+        """
         # Note: This might go away once we clean up the mapping logic server-side.
         iq.result.label = convert_internal_label_to_display(iq, iq.result.label)
+        if want_async:
```
robotrapta, line 82: Nice and clean.

src/groundlight/client.py
```diff
@@ -214,6 +222,15 @@ def submit_image_query(  # noqa: PLR0913 # pylint: disable=too-many-arguments
...
             params["human_review"] = human_review
 
+        if want_async is True:
+            # If want_async is True, we don't want to wait for a result. As a result wait must be set to 0 to use
+            # want_async.
+            if wait != 0:
+                raise ValueError(
+                    "wait must be set to 0 to use want_async. Using wait and want_async at the same time is incompatible."  # noqa: E501
+                )
+            params["want_async"] = want_async
```
robotrapta, line 232: I think here we should be more deliberate about what we send to the API.  We can be pretty flexible in what the SDK method uses to indicate that they want async - they can give us an actual bool, or a 0/1 int or something else which indicates truthiness.  But we should be more specific about what we send to the server.  In the other PR I argue for always sending “1” or “0”.

src/groundlight/client.py
```diff
@@ -225,10 +242,27 @@ def submit_image_query(  # noqa: PLR0913 # pylint: disable=too-many-arguments
...
+        detector: Union[Detector, str],
+        image: Union[str, bytes, Image.Image, BytesIO, BufferedReader, np.ndarray],
+        human_review: Optional[str] = None,
+        inspection_id: Optional[str] = None,
+    ) -> ImageQuery:
+        """
+        Convenience method for submitting an image query asynchronously. This is equivalent to calling
+        submit_image_query with want_async=True and wait=0. Use get_image_query to retrieve the result of the image
+        query.
+        """
```
robotrapta, line 262: We should flesh this out with full documentation of all the parameters accepted.

I have a bunch of thoughts of how to word the summary here, but I’d rather type them when I’m at a proper keyboard.

robotrapta (approved):
This LGTM - some suggestions within, and of course we should be conscious of dependencies when merging.

Also, this probably deserves a major version bump to 0.12, especially with the other `ask_` methods.

.vscode/settings.json
```diff
@@ -3,6 +3,17 @@
   "editor.codeActionsOnSave": {
     "source.organizeImports": true
   },
-  "python.analysis.extraPaths": ["./generated"],
-  "python.formatting.provider": "black"
-}
+  "editor.rulers": [
```
blaise-muhirwa, line 6: Looks like you accidentally committed this!

sunildkumar: this was intentional - I was just adding the stuff we have in zuuul that we needed here to
1. make it easier to keep code linted
2. make black behave rationally

spec/public-api.yaml
sunildkumar: Talked to mike about this - we discussed the aspirational goal of organizational/behavioral change to ensure this stays up to date.

src/groundlight/client.py
sunildkumar: with the serverside change (where the serve returns result=None, this is no longer necessary, so it's even cleaner!

src/groundlight/client.py
sunildkumar: Why not just True/False?

src/groundlight/client.py
sunildkumar: I copy pasted the relevant descriptions from submit_image_query for now.

.vscode/settings.json
sunildkumar: ...at least on my machine. the rulers are helpful to make the linter happy and black does wild things with the imports if it is allowed to touch them.

src/groundlight/client.py
brandon-groundlight: I'd defer to Leo. My understanding is the the "1" "0" distinction fixes the bits representation. True values can have multiple representations (that occasionally are different by system) so we might as well choose a 1:1 mapping on the values sent into our system rather than a many to one mapping. 
Again, I defer to Leo

brandon-groundlight (approved):

src/groundlight/client.py
sunildkumar: I disagree with this pretty strongly as we are going to essentially going to have to write a two translation layers (on both the SDK and BE) to convert to and from these "string bools". OTOH, python's `True` and `False` are already unambiguous values for T/F. Additionally the SDK automatically handles serialization of `True`/`False` and our BE automatically handles deserialization to `True`/`False` so no custom translation is required.

sunildkumar: If we really want this, I think we should do it as its own PR as that way we can focus on applying this update uniformly as there is BE work required as well.

sunildkumar: If we want to accept other arguments to represent truthiness from the user I think the solution is to do validation on the inputs prior to whatever business logic we do.

src/groundlight/client.py
```diff
@@ -336,6 +340,15 @@ def submit_image_query(  # noqa: PLR0913 # pylint: disable=too-many-arguments
         if human_review is not None:
             params["human_review"] = human_review
 
+        if want_async is True:
```
blaise-muhirwa, line 343:
```suggestion
        if want_async:
```

blaise-muhirwa (approved):
Awesome! As far as documenting this with sphinx, it won't be a problem since sphinx autogenerates docs for everything under `client.py`. Just to make sure things look good tho you can run `make develop-docs-comprehensive` to make sure the docs look good!

src/groundlight/client.py
robotrapta:
```suggestion
            params["want_async"] = str(bool(want_async))
```
is what I mean.  Don't rely on the variable's python-truthiness and its string representation being the same.

PR #104, opened 2023-10-10
added review_reasons to the SDK for posting labels

sunildkumar:
Serverside will soon strongly encourage that a `review_reason` is posted with a label (this value represents why the label was created). For now, all labels created by the SDK are `CUSTOMER_INITIATED` so all we need to do is post this value whenever a SDK user creates a label. 

Blockers:
~~1. PR feedback~~
2. The BE PR should be merged first, as otherwise the SDK will be posting data the server isn't capable of handling
3. Tests must pass (right now everything but detector improvement passes and I've run it a few times...). 
4. Personally I would like to push a package update to pypi so our customers' SDK labels have the appropriate field in the POST rather than relying on the server to pick a sensible default.
5. BUMP TOML


paulina-positronix (approved):

generated/model.py
```diff
@@ -84,3 +84,10 @@ class PaginatedImageQueryList(BaseModel):
     next: Optional[AnyUrl] = Field(None, example="http://api.example.org/accounts/?page=4")
     previous: Optional[AnyUrl] = Field(None, example="http://api.example.org/accounts/?page=2")
     results: Optional[List[ImageQuery]] = None
+
+# These are reasons a label was created. A review reason is a required field when posting a human label to the API.
+class ReviewReason(str, Enum):  # noqa: N801
+    ESCALATION = "ESCALATION"
+    AUDIT = "AUDIT"
+    ASSESSMENT = "ASSESSMENT"
+    CUSTOMER_INITIATED = "CUSTOMER_INITIATED"
```
mjvogelsong, line 93: A SDK user should only be setting `CUSTOMER_INITIATED`. The other reasons shouldn't be used for submission, so I don't think we should include them here.

(We could add them later if we return reasons in the API response, but for now, I think it would just cause confusion).

mjvogelsong: Also, I don't think this should be in `generated/` since it's not auto-generated (and the auto-generator is gonna overwrite it next time we run it). Instead, it should be in a file under `src/groundlight/` -- I think it would be okay to put in `internalapi.py`.

mjvogelsong (commented):
Let's bump the pyproject.toml minor version as well!

generated/model.py
sunildkumar: done!

mjvogelsong (approved):

PR #105, opened 2023-10-10
Update UNSURE to be UNCLEAR, catch __USURE received from the service

brandon-groundlight:
Updates the labels displayed by the SDK to be YES, NO, and UNCLEAR


sunildkumar (approved):
Surprisingly, I think I would classify this as a breaking change to the SDK, as our customers are likely to be doing something with the string values we emit. So we probably need to do a version bump here. Otherwise it looks fine with me.

PR #108, opened 2023-10-11
Api ref docs deployment

blaise-muhirwa:


.github/workflows/cicd.yaml
```diff
@@ -83,6 +83,33 @@ jobs:
       - name: Build website
         run: npm run build
 
+    # Checks that the API reference docs built with shpinx build
+  test-api-reference-docs:
```
blaise-muhirwa, line 87: Moved the previous `sphinx_docs.yaml` here. Now we're also building the docs on every PR.

docs/docusaurus.config.js
```diff
@@ -24,8 +24,8 @@ const config = {
   deploymentBranch: "gh-pages", // Branch that GitHub pages will deploy from.
   trailingSlash: false,
 
-  onBrokenLinks: "throw",
-  onBrokenMarkdownLinks: "warn",
+  onBrokenLinks: "ignore",
```
blaise-muhirwa, line 27: These shouldn't be ignored, but for some reason docusaurus keeps complaining about `docs/api-reference` being a broken link. Nonetheless, this builds correctly and the docs page is properly rendered in the browser.

docs/api-reference/api-reference.md
robotrapta, line 1: I don't think this page gets rendered at all right now.

.github/workflows/cicd.yaml
```diff
@@ -204,6 +232,18 @@ jobs:
           cache: npm
       - name: Install dependencies
         run: npm install
+      - name: Download api-reference docs 
+        # download docs to python-sdk/api-reference-docs
+        uses: actions/download-artifact@v2
```
robotrapta, line 237: This is a nice trick, but relying on github action tools makes it difficult to test locally.  We should put all this into a bash script or make target so we can preview a full docs build.

docs/docusaurus.config.js
robotrapta: It's complaining because the URL `/python-sdk/docs/api-reference` doesn't exist.  `api-reference` is one level up from the other topics, so it's at `/python-sdk/api-reference`.

robotrapta:
```suggestion
  onBrokenLinks: "throw",
```
This needs to stay on.

docs/docusaurus.config.js
```diff
@@ -24,8 +24,8 @@ const config = {
...
+  onBrokenMarkdownLinks: "ignore",
```
robotrapta, line 28: These tests are very valuable for catching docs mistakes.  Without this test, we'll get lots of docs bugs.  I'm surprised `onBrokenMarkdownLinks` wasn't configured for failure before - we should actually turn that back on.

docs/docusaurus.config.js
```diff
@@ -123,6 +123,10 @@ const config = {
                 label: "Installation",
                 to: "/docs/installation",
               },
+              {
+                label: "API Reference",
+                to: "/docs/api-reference",
```
robotrapta, line 128:
```suggestion
                to: "/api-reference-docs/",
```
This would make it link to the location it's currently building to, if that's where we want to keep it.

docs/api-reference/_category_.json
robotrapta, line 1: I think this file & its `.md` are both in the wrong directory.  They should be under `docs/docs`.

robotrapta (changes requested):
Glad to see them working together.  It would be nice to have the API reference fully integrated with docusaurus's table-of-contents, but I'm guessing that'll take writing some code.  And just including the link like this is a big step in the right direction.

But the links are definitely not right as is.  The footer link is broken, and I don't think there's a TOC link at all.  I'd root-cause this mistake as being caused by the doc build relying on github actions.  We should move the docs build process to something that can be fully tested locally - one script or Makefile-target that you can run which builds the sphinx docs and the docusaurus docs together so we can see what it will all look like together.

Another thing to put on the TODO list would be moving to vercel for hosting - they make it easy to preview branches.  That's also probably more than we want to take on right now.  This is quite close, and we should finish it up, because it's a huge improvement!

.github/workflows/cicd.yaml
blaise-muhirwa: Added a make target for this: `make docs-comprehensive`

docs/docusaurus.config.js
```diff
@@ -98,6 +98,11 @@ const config = {
             label: "Applications",
             position: "left",
           },
+          {
+            href: "pathname:///python-sdk/api-reference-docs/",
```
blaise-muhirwa, line 102: Also `pathname://` here is required in order for the sphinx docs to be properly rendered without having to refresh the page. For more on this: https://docusaurus.io/docs/advanced/routing#escaping-from-spa-redirects

Makefile
```diff
@@ -72,12 +72,25 @@ sphinx-help:
 # The .PHONY directive tells make that `apidocs` and `html` are labels for 
 # commands. `apidocs: html` allows us to generate docs by running 
 # `make apidocs` instead. 
-.PHONY: apidocs html 
-
+.PHONY: docs-comprehensive apidocs html 
+
+# Start an interactive server to test docs locally
+test-docs-comprehensive: docs-comprehensive
```
robotrapta, line 78:
```suggestion
develop-docs-comprehensive: docs-comprehensive 
```
The term "test" implies it's good as part of an automated test suite, which `npm start` is not.  So I'd call it develop.

docs/README.md
```diff
@@ -9,9 +9,7 @@ The docs are included with the SDK so that we can automate testing of the code s
 Doc changes are published automatically when they're merged to main.  To preview changes, build and host the site locally.  You'll need a reasonably modern version of `npm` and then:
 
 ```
-npm install
-npm run build  # looks for any errors
-npm start  # starts interactive server
+make test-docs-comprehensive
```
robotrapta, line 12:
```suggestion
make develop-docs-comprehensive
```
This is pretty nice.

docs/package.json
```diff
@@ -17,7 +17,7 @@
   "dependencies": {
     "@docusaurus/core": "2.4.0",
     "@docusaurus/preset-classic": "2.4.0",
-    "@easyops-cn/docusaurus-search-local": "^0.35.0",
+    "@easyops-cn/docusaurus-search-local": "^0.29.0",
```
robotrapta, line 20: Why did we downgrade?

robotrapta (approved):
Awesome

docs/package.json
blaise-muhirwa: This is not required anymore! Earlier when I had a broken link, the downgrade somehow made it possible for docusaurus to still be able to find the sphinx docs.

PR #109, opened 2023-10-11
Cli polishing

brandon-groundlight:
Polishing up the behavior of the CLI, granting access to the help tools before specifying an API token and raising the API token error only when trying to execute a command.


src/groundlight/cli.py
```diff
@@ -14,9 +16,19 @@ def class_func_to_cli(method):
     Given the class method, simplify the typing on the inputs so that Typer can accept the method
     """
 
-    @wraps(method)
+    # We create a temporary class to bind the method to so we have the correct annotations for typer to use
+    # The first thing we'll do when we actually call the method is to bind it to a proper Groundlight instance
```
brandon-groundlight, line 20: This is some of the wildest python I've ever written. My conclusion is that if we want to make any significant changes to this behavior in the future we immediately ditch Typer and write our own argparser. Sans that, this does seem to do what we want it to do.

test/unit/test_cli.py
```diff
@@ -72,12 +84,18 @@ def test_detector_and_image_queries():
     assert completed_process.returncode == 0
 
 
-def test_help():
+def test_help(no_api_token):
```
mjvogelsong, line 87: Not a big deal, but a slightly simpler alternative to the fixture approach would be to use `mock.patch.dict()`

```python
import os
from unittest.mock import patch

@patch.dict(os.environ, {'GROUNDLIGHT_API_TOKEN': None})
def test_help():
```

src/groundlight/cli.py
```diff
@@ -5,18 +5,33 @@
...
 
 def class_func_to_cli(method):
     """
-    Given the class method, simplify the typing on the inputs so that Typer can accept the method
+    Given the class method, create a method with the identical signature to provide the help documentation and
+    but only instantiates the class when the method is actually called.
     """
 
-    @wraps(method)
+    # We create a fake class and fake method to so we have the correct annotations for typer to use
```
mjvogelsong, line 20:
```suggestion
    # We create a fake class and fake method so we have the correct annotations for typer to use
```

src/groundlight/cli.py
mjvogelsong, line 1: On the help message, it truncates the descriptions of the sub-commands. Is there a way to have it print out the full line?

```
Commands:
  add-label                       Add a new label to an image query.
  create-detector                 Create a new detector with a...
  get-detector                    Get a detector by id
  get-detector-by-name            Get a detector by name
  get-image-query                 Get an image query by id
  get-or-create-detector          Tries to look up the detector by...
  list-detectors                  List out detectors you own
  list-image-queries              List out image queries you own
  start-inspection                For users with Inspection Reports...
  stop-inspection                 For users with Inspection Reports...
  submit-image-query              Evaluates an image with Groundlight.
  update-detector-confidence-threshold
                                  Updates the confidence threshold...
  update-inspection-metadata      For users with Inspection Reports...
  wait-for-confident-result       Waits for an image query result's...
```

src/groundlight/config.py
```diff
@@ -8,3 +8,9 @@
     "API_TOKEN_VARIABLE_NAME",
     "DEFAULT_ENDPOINT",
 ]
+
+API_TOKEN_HELP_MESSAGE = (
+    "No API token found. Please put your token in an environment variable "
+    f'named "{API_TOKEN_VARIABLE_NAME}". If you don\'t have a token, you can '
+    f"create one at {API_TOKEN_WEB_URL}"
+)
```
sunildkumar, line 16: great idea to standardize this

sunildkumar (approved):

src/groundlight/cli.py
mjvogelsong: On the help page for a specific command, the description's formatting can be a little weird because it doesn't respect *single* newlines. (I understand that `typer.Argument()` would fix this if we could use it, but it doesn't work because of `Union` types).

```
❯ poetry run groundlight add-label -h
Usage: groundlight add-label [OPTIONS] IMAGE_QUERY LABEL

  Add a new label to an image query.  This answers the detector's question.
  :param image_query: Either an ImageQuery object (returned from
  `submit_image_query`) or an image_query id as a string. :param label: The
  string "YES" or the string "NO" in answer to the query.

Arguments:
  IMAGE_QUERY  [required]
  LABEL        [required]

Options:
  -h, --help  Show this message and exit.
```

But I think the fix is actually pretty easy -- when we want a newline to show up in the generated message, we should just put in a blank line in the docstring. For example:

```python
    def add_label(self, image_query: Union[ImageQuery, str], label: Union[Label, str]):
        """Add a new label to an image query.  This answers the detector's question.

        :param image_query: Either an ImageQuery object (returned from `submit_image_query`) or
        an image_query id as a string.

        :param label: The string "YES" or the string "NO" in answer to the query.
        """
```

Which then looks better:

```
poetry run groundlight add-label -h
Usage: groundlight add-label [OPTIONS] IMAGE_QUERY LABEL

  Add a new label to an image query.  This answers the detector's question.

  :param image_query: Either an ImageQuery object (returned from
  `submit_image_query`) or an image_query id as a string.

  :param label: The string "YES" or the string "NO" in answer to the query.
```

mjvogelsong (approved):
Cool, nice updates!

src/groundlight/cli.py
brandon-groundlight: The docstrings are undergoing continual improvement, they'll all hopefully use the double new line once we finish working out the sphinx stuff.

Turns out not truncating lines was just a matter of adding a param to the context_settings of the typer app.

PR #110, opened 2023-10-12
Don't run create_detector doc test because it is not repeatable

tyler-romero:
Fixed bug on backend that makes this test *correctly* fail when run repeatedly (because we dont allow duplicate predictor names in the same predictor_set). We still want to demonstrate this method though, so we just dont run the doc test.


Makefile
```diff
@@ -42,6 +42,9 @@ test-integ: install  ## Run tests against the integ API server (needs GROUNDLIGH
 test-docs: install  ## Run the example code and tests in our docs against the prod API (needs GROUNDLIGHT_API_TOKEN)
 	poetry run pytest -v --markdown-docs ${TEST_ARGS} docs README.md
 
+test-docs-integ: install  ## Run the example code and tests in our docs against the prod API (needs GROUNDLIGHT_API_TOKEN)
```
tyler-romero, line 45:
```suggestion
test-docs-integ: install  ## Run the example code and tests in our docs against the integ API (needs GROUNDLIGHT_API_TOKEN)
```

mjvogelsong (approved):

PR #111, opened 2023-10-12
Add Flag to Disable SSL/TLS Certificates Verification 

blaise-muhirwa:
Adds a constructor parameter called `disable_tls_verification` and optionally allows users to disable SSL/TLS verification via an env var called `TLS_VERIFY`. 

I tested this on the edge endpoint where I generated self-signed TLS certificates with OpenSSL and then used the SDK from this branch to check that sending requests to an HTTPS server running on local host still works.


src/groundlight/client.py
```diff
@@ -95,6 +106,20 @@ def __init__(self, endpoint: Optional[str] = None, api_token: Optional[str] = No
...
+        should_disable_tls = disable_tls_verification or os.environ.get(DISABLE_TLS_VARIABLE_NAME, None) is not None
+
+        if should_disable_tls:
+            logger.info("Disabling SSL/TLS certificate verification")
+
+            import warnings
+
+            from urllib3.exceptions import InsecureRequestWarning
+
+            warnings.simplefilter("ignore", InsecureRequestWarning)
```
blaise-muhirwa, line 118: This suppresses the following warning: 

```
/home/ubuntu/.cache/pypoetry/virtualenvs/groundlight-edge-ZbDJXDWz-py3.10/lib/python3.10/site-packages/urllib3/connectionpool.py:1056: InsecureRequestWarning: Unverified HTTPS request is being made to host 'localhost'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/1.26.x/advanced-usage.html#ssl-warnings
  warnings.warn(
```

src/groundlight/client.py
```diff
@@ -95,6 +106,20 @@ def __init__(self, endpoint: Optional[str] = None, api_token: Optional[str] = No
                     ),
                 ) from e
 
+        should_disable_tls = disable_tls_verification or os.environ.get(DISABLE_TLS_VARIABLE_NAME, None) is not None
+
+        if should_disable_tls:
+            logger.info("Disabling SSL/TLS certificate verification")
```
blaise-muhirwa, line 112: Should this be a `logger.warn`?

src/groundlight/client.py
```diff
@@ -64,7 +64,12 @@ class Groundlight:
     POLLING_EXPONENTIAL_BACKOFF = 1.3  # This still has the nice backoff property that the max number of requests
     # is O(log(time)), but with 1.3 the guarantee is that the call will return no more than 30% late
 
-    def __init__(self, endpoint: Optional[str] = None, api_token: Optional[str] = None) -> None:
+    def __init__(
+        self,
+        endpoint: Optional[str] = None,
+        api_token: Optional[str] = None,
+        disable_tls_verification: Optional[bool] = False,
```
robotrapta, line 71:
```suggestion
        disable_tls_verification: Optional[bool] = None,
```
I think we should follow the same precedence logic we use for the `API_TOKEN` -- if it's set in code, use that; if it's not set in code, check the environment variable.  If we want to follow that logic, then we need the default here to be `None` so we can tell the difference between the parameter being set or not.

src/groundlight/client.py
```diff
@@ -75,6 +80,12 @@ def __init__(self, endpoint: Optional[str] = None, api_token: Optional[str] = No
                         If unset, fallback to the environment variable "GROUNDLIGHT_API_TOKEN".
         :type api_token: str
 
+        :param disable_tls_verification: Set this to false to skip verifying SSL/TLS certificate
+                                        when calling API from https server. In unset, fallback to
+                                        the environment variable "TLS_VERIFY". By default, certificates
+                                        are verified.
```
robotrapta, line 86:
```suggestion
        :param disable_tls_verification: Set this to `True` to skip verifying SSL/TLS certificates
                                        when calling API from https server. In unset, fallback it will check
                                        the environment variable "TLS_VERIFY" for `1` or `0`. By default, 
                                        certificates are verified.

                                        Disable verification if using a self-signed TLS certificate with a Groundlight 
                                        Edge Endpoint.  It is unadvised tto disable verification  if connecting directly 
                                        to the Groundlight cloud service. 
```

src/groundlight/client.py
```diff
@@ -75,6 +80,12 @@ def __init__(self, endpoint: Optional[str] = None, api_token: Optional[str] = No
...
+        :type disable_tls_verification: bool
```
robotrapta, line 87:
```suggestion
        :type disable_tls_verification: Optional[bool]
```

src/groundlight/client.py
```diff
@@ -95,6 +106,20 @@ def __init__(self, endpoint: Optional[str] = None, api_token: Optional[str] = No
                     ),
                 ) from e
 
+        should_disable_tls = disable_tls_verification or os.environ.get(DISABLE_TLS_VARIABLE_NAME, None) is not None
```
robotrapta, line 109: This logic isn't quite right.  To be consistent with the API_TOKEN we should check if the specified variable is `None` and if so, read the environment variable.

Environment variables are strings, so we should parse them with `bool(int(os.environ.get(DISABLE_TLS_VARIABLE_NAME, 0)))`.  If we just interpret the string as a bool, then any string will be considered `True`, even the string "no" or "False" or "0".

src/groundlight/client.py
```diff
@@ -95,6 +106,20 @@ def __init__(self, endpoint: Optional[str] = None, api_token: Optional[str] = No
...
+
+        if should_disable_tls:
```
robotrapta, line 111:
```suggestion
        if should_disable_tls_verification:
```
We're not really disabling TLS - just the verification.

src/groundlight/client.py
robotrapta: I think warn is a good idea.

robotrapta:
```suggestion
            logger.warning("Disabling SSL/TLS certificate verification.  This should only be used when connecting to an endpoint with a self-signed certificate.")
```

src/groundlight/client.py
```diff
@@ -95,6 +106,20 @@ def __init__(self, endpoint: Optional[str] = None, api_token: Optional[str] = No
...
+            logger.info("Disabling SSL/TLS certificate verification")
+
+            import warnings
```
robotrapta, line 114: Let's keep the imports at the top of the file unless we really can't.

robotrapta (commented):
Nice.  A couple changes suggested for precedence logic and picking the configuration.

src/groundlight/client.py
```diff
@@ -95,6 +112,20 @@ def __init__(self, endpoint: Optional[str] = None, api_token: Optional[str] = No
                     ),
                 ) from e
 
+        should_disable_tls_verification = disable_tls_verification or bool(
```
robotrapta, line 115: Still not right.  We want the parameter's value in code, if set, to take precedence over the environment variable.  Here, if the code sets the value to `False` but the environment variable is true (`1`) then the result will be true.

We should only look at the environment variable if the parameter is `None`.

robotrapta (commented):
Almost there.

robotrapta (approved):
Nice!  Sorry for the slow review.

PR #117, opened 2023-10-13
new tool for running the docs server

sunildkumar:
Tool co-authored with chatgpt to auto-magically update the docs server when you update the code in this directory. 

There is some process killing in the script so IDK how safe it is. GPT ensured me it was ok...so I'm only slightly less terrified.

### DIFF
--- a/start_docs_server.sh
+++ b/start_docs_server.sh
@@ -1,46 +1,48 @@
 #!/bin/bash
 # convient script to run the docs server. It automatically rebuilds and restarts when you change code. You just need to refresh your browser. 
+cd "$(dirname "$0")"
 WATCH_PATH="."
 START_SERVER_CMD="make develop-docs-comprehensive"
-
 while true; do
   $START_SERVER_CMD &
-  SERVER_PID=$!
   
-  echo "Server started with PID: $SERVER_PID"
+  echo "Server started"
 
   echo "Waiting for 15 seconds before starting to watch for file changes..."
   sleep 15
   
   if [[ "$OSTYPE" == "darwin"* ]]; then
-    fswatch -1 --exclude 'docs/static/api-reference-docs' --exclude 'build/' --exclude '/docs/.docusaurus' --exclude 'changes.log' --exclude 'docs/node_modules/.cache/webpack' --exclude '.git/'  $WATCH_PATH | tee -a changes.log
+    CHANGED_FILE=$(fswatch -1 --exclude 'docs/static/api-reference-docs' --exclude 'build/' --exclude '/docs/.docusaurus' --exclude 'changes.log' --exclude 'docs/node_modules/.cache/webpack' --exclude '.git/'  $WATCH_PATH)
+    echo "Detected changes in: $CHANGED_FILE"
   else
     echo "OS not supported"
     exit 1
   fi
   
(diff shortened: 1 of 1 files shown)


start_docs_server.sh
```diff
@@ -0,0 +1,46 @@
...
+  sleep 15
+  
+  if [[ "$OSTYPE" == "darwin"* ]]; then
+    fswatch -1 --exclude 'docs/static/api-reference-docs' --exclude 'build/' --exclude '/docs/.docusaurus' --exclude 'changes.log' --exclude 'docs/node_modules/.cache/webpack' --exclude '.git/'  $WATCH_PATH | tee -a changes.log
+  else
+    echo "OS not supported"
+    exit 1
+  fi
+  
+  echo "Code changed. Attempting to kill server with PID: $SERVER_PID..."
```
robotrapta, line 22: Well, you're not actually.  You're killing `npm` by name.

start_docs_server.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+# convient script to run the docs server. It automatically rebuilds and restarts when you change code. You just need to refresh your browser. 
+WATCH_PATH="."
+START_SERVER_CMD="make develop-docs-comprehensive"
+
+while true; do
+  $START_SERVER_CMD &
+  SERVER_PID=$!
```
robotrapta, line 8: I would guess this is not capturing the PID of the web server, but rather the `make` process, which launches the web server as a sub-process.  Which is maybe why you don't use this later?

start_docs_server.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+# convient script to run the docs server. It automatically rebuilds and restarts when you change code. You just need to refresh your browser. 
+WATCH_PATH="."
```
robotrapta, line 3: Maybe wanna add `cd "$(dirname "$0")"` to the top to make sure the script is running from where you expect it to.  If people say "../../start_docs_server.sh" for example this wouldn't work super well.

robotrapta (approved):
Happy to have this tool in the toolbox.  Some comments look wrong, and I"m sure it could be improved.  But getting this script perfect is not at all important.  Happy to take it as is too.

start_docs_server.sh
sunildkumar: Agreed. Trying to kill via PID actually left stranded processes running that kept the old version of the server up. IDK why GPT didn't change the comments...

start_docs_server.sh
sunildkumar: yes exactly

start_docs_server.sh
sunildkumar: good idea

start_docs_server.sh
```diff
@@ -0,0 +1,46 @@
...
+      exit 1
+    fi
+  fi
+  
+  # Additional sleep to ensure port is released before restarting the server.
+  echo "Waiting for an additional 5 seconds before restarting server..."
+  sleep 5
+  
+  # Clear the log file to avoid reading old changes
+  > changes.log
```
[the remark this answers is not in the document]
sunildkumar: I removed this part and now instead dump directly to bash when it detects changes

PR #122, opened 2023-10-17
Disable improvement test

brandon-groundlight:
Temporarily disable tests dependent on code external to the SDK

### DIFF
--- a/test/integration/test_groundlight.py
+++ b/test/integration/test_groundlight.py
@@ -9,7 +9,7 @@ import openapi_client
 import pytest
 from groundlight import Groundlight
 from groundlight.binary_labels import VALID_DISPLAY_LABELS, DeprecatedLabel, Label, convert_internal_label_to_display
-from groundlight.internalapi import InternalApiError, NotFoundError
+from groundlight.internalapi import InternalApiError, NotFoundError, iq_is_answered
 from groundlight.optional_imports import *
 from groundlight.status_codes import is_user_error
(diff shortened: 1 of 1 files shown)


test/integration/test_groundlight.py
```diff
@@ -456,86 +456,88 @@ def test_submit_numpy_image(gl: Groundlight, detector: Detector):
...
-    assert iq_is_answered(fast_always_yes_iq)
-    name = f"Test {datetime.utcnow()}"  # Need a unique name
-    query = "Is there a dog?"
-    detector = gl.create_detector(name=name, query=query, confidence_threshold=0.8)
-    fast_iq = gl.ask_ml(detector=detector.id, image="test/assets/dog.jpeg", wait=0)
-    assert iq_is_answered(fast_iq)
-    confident_iq = gl.ask_confident(detector=detector.id, image="test/assets/dog.jpeg", wait=180)
-    assert confident_iq.result.confidence > IQ_IMPROVEMENT_THRESHOLD
+# temporarily disabled until the backend bugfix is deployed
+# @pytest.mark.skipif(MISSING_PIL, reason="Needs pillow")  # type: ignore
```
sunildkumar, line 460: Instead of commenting out you might just use `@pytest.mark.skip(reason="Detector improvement is incredibly flaky")` or something like that?

sunildkumar (approved):
I think it would be better to just use the decorator instead of commenting out.

PR #125, opened 2023-10-18
Improve the plumbing and documentation for some of the ask* methods

tyler-romero:
Address a some documentation papercuts for the new ask_* methods. Plus, improve plumbing of arguments to `submit_image_query` and to `wait_for_confident_result`.

### DIFF
--- a/docs/docs/building-applications/5-async-queries.md
+++ b/docs/docs/building-applications/5-async-queries.md
@@ -16,7 +16,7 @@ from time import sleep
 
 detector = gl.get_or_create_detector(name="your_detector_name", query="your_query")
 
-cam = cv2.VideoCapture(0)  # Initialize camera (0 is the default index) 
+cam = cv2.VideoCapture(0)  # Initialize camera (0 is the default index)
 
 while True:
     _, image = cam.read()  # Capture one frame from the camera
@@ -35,7 +35,7 @@ from groundlight import Groundlight
 
 detector = gl.get_or_create_detector(name="your_detector_name", query="your_query")
 
-image_query_id = db.get_next_image_query_id()  
+image_query_id = db.get_next_image_query_id()
 
 while image_query_id is not None:
     image_query = gl.get_image_query(id=image_query_id) # retrieve the image query from Groundlight
@@ -67,4 +67,7 @@ result = image_query.result  # This will always be 'None' as you asked asynchron
 
 image_query = gl.get_image_query(id=image_query.id)  # Immediately retrieve the image query from Groundlight
 result = image_query.result  # This will likely be 'UNCLEAR' as Groundlight is still processing your query
+
+image_query = gl.wait_for_confident_result(id=image_query.id)  # Poll for a confident result from Groundlight
+result = image_query.result
 ```
--- a/src/groundlight/client.py
+++ b/src/groundlight/client.py
@@ -428,9 +428,10 @@ class Groundlight:
         detector: Union[Detector, str],
         image: Union[str, bytes, Image.Image, BytesIO, BufferedReader, np.ndarray],
         wait: Optional[float] = None,
-        # human_review: Optional[str] = None,
     ) -> ImageQuery:
-        """Evaluates an image with Groundlight, getting the first answer Groundlight can provide.
+        """
+        Evaluates an image with Groundlight, getting the first answer Groundlight can provide.
+
         :param detector: the Detector object, or string id of a detector like `det_12345`
         :type detector: Detector or str
 
--- a/src/groundlight/client.py
+++ b/src/groundlight/client.py
@@ -1,3 +1,4 @@
+from functools import partial
 import logging
 import os
 import time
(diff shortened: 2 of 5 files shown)


src/groundlight/client.py
```diff
@@ -375,8 +387,10 @@ def ask_confident(
         confidence_threshold: Optional[float] = None,
         wait: Optional[float] = None,
     ) -> ImageQuery:
-        """Evaluates an image with Groundlight waiting until an answer above the confidence threshold
-            of the detector is reached or the wait period has passed.
+        """
+        Evaluates an image with Groundlight waiting until an answer above the confidence threshold
```
tyler-romero, line 391: Address some formatting issues
https://code.groundlight.ai/python-sdk/api-reference-docs/#groundlight.client.Groundlight.ask_confident

src/groundlight/client.py
```diff
@@ -469,6 +488,17 @@ def ask_async(
...
+            The longer the patience_time, the more likely Groundlight will arrive at a confident answer.
+            Within patience_time, Groundlight will update ML predictions based on stronger findings,
+            and, additionally, Groundlight will prioritize human review of the image query if necessary.
+            This is a soft server-side timeout. If not set, use the detector's patience_time.
+        :type patience_time: float
+
+        :param confidence_threshold: The confidence threshold to wait for.
+            If not set, use the detector's confidence threshold.
+        :type confidence_threshold: float
+
```
sunildkumar, line 501: I absolutely love this description of patience time

src/groundlight/client.py
```diff
@@ -500,26 +530,35 @@ def ask_async(
...
+            # retrieve the result later or on another machine by calling gl.wait_for_confident_result()
+            # with the id of the image_query above. This will block until the result is available.
+            image_query = gl.wait_for_confident_result(image_query.id)
 
             # now the result will be available for your use
             assert image_query.result is not None
 
+            # alternatively, you can check if the result is available (without blocking) by calling
+            # gl.get_image_query() with the id of the image_query above. This will return the image_query,
+            # but the result may still be None.
```
sunildkumar, line 545: The result is only None on the servers response to the original ask_async query. At worst the result here is `UNCLEAR` (this is if the best_prediction is None or the placeholder).

src/groundlight/client.py
```diff
@@ -538,10 +578,12 @@ def wait_for_confident_result(
         :return: ImageQuery
         :rtype: ImageQuery
         """
+        if confidence_threshold is None:
+            if isinstance(image_query, str):
+                image_query = self.get_image_query(image_query)
+            confidence_threshold = self.get_detector(image_query.detector_id).confidence_threshold
```
sunildkumar, line 584: my only concern here is that this introduces another request from the server. I recall @robotrapta suggesting that we should try to minimize this to minimize overall latency, but maybe I am mistaken. I am guessing that the alternative would be to force the user to provide the detector as an argument to the function which is kinda ugly.

src/groundlight/client.py
```diff
@@ -538,10 +578,12 @@ def wait_for_confident_result(
...
 
-        def confidence_above_thresh(iq):
-            return iq_is_confident(iq, confidence_threshold=confidence_threshold)
-
+        confidence_above_thresh = partial(iq_is_confident, confidence_threshold=confidence_threshold)
```
sunildkumar, line 586: thanks for teaching me about `partial` - this is very haskell-y.

src/groundlight/client.py
```diff
@@ -623,6 +662,7 @@ def add_label(self, image_query: Union[ImageQuery, str], label: Union[Label, str
         else:
             image_query_id = str(image_query)
             # Some old imagequery id's started with "chk_"
+            # TODO: handle iqe
```
sunildkumar, line 665: I'm not sure what you mean here, so I'm concerned that others might not as well, but maybe I'm just out of the loop.

docs/docs/building-applications/5-async-queries.md
```diff
@@ -67,4 +67,7 @@ result = image_query.result  # This will always be 'None' as you asked asynchron
 
 image_query = gl.get_image_query(id=image_query.id)  # Immediately retrieve the image query from Groundlight
 result = image_query.result  # This will likely be 'UNCLEAR' as Groundlight is still processing your query
+
+image_query = gl.wait_for_confident_result(id=image_query.id)  # Poll for a confident result from Groundlight
+result = image_query.result
```
sunildkumar, line 72: I like showing this as the default use case.

src/groundlight/client.py
tyler-romero: We actually call get_image_query if image_query is a string as the very first step in `_wait_for_result`, so this just brings the call forwards a bit.

src/groundlight/client.py
sunildkumar, line 545:
```suggestion
            # but the result may still be UNCLEAR.
```
possible fix...but not the only option

src/groundlight/client.py
tyler-romero: Haha its honestly really nice sometimes. I love it because its a bit of a functional holdover

sunildkumar (approved):
nice! Thanks for your feedback on all of the documentation this afternoon.

src/groundlight/client.py
tyler-romero: Edge detectors return image queries with the prefix `iqe_`, so we need to work on a way to handle that sanely

src/groundlight/client.py
sunildkumar: ah ok nice

src/groundlight/client.py
brandon-groundlight: This what the edge server will return, right? This just opens up the sdk to work with the edge

sunildkumar: got it!

brandon-groundlight (approved):
Did we also want to attach confidence_threshold to the object sent to the BE, or does that require some BE work first?

src/groundlight/client.py
tyler-romero: Exactly Brandon, just updated the comment to be a bit more clear

tyler-romero:
That will require a bit of BE work Brandon, this is just the intermediate solution

PR #131, opened 2023-11-07
blocking tests getting in the way of CICD

brandon-groundlight:
While useful tests of our SLA, these tests when failing provide little actionable choices and only an alert on system performance.


tomfaulhaber (approved):
Looks good

PR #132, opened 2023-11-09
Add IQ metadata

mjvogelsong:
Allow SDK users to add a small `metadata` field to `submit_image_query()` and `ask_*()` methods.


sunildkumar (approved):
looks good!

PR #135, opened 2023-11-10
(small) Add SDK version to API requests

mjvogelsong:
This makes it a little easier to debug customer SDK issues when they're using an older SDK version.


src/groundlight/__init__.py
```diff
@@ -5,13 +5,6 @@
 # Imports from our code
 from .client import Groundlight
 from .binary_labels import Label
+from .version import get_version
 
-try:
```
mjvogelsong, line 9: I had to move this to avoid a circular dependency.

tyler-romero (approved):

PR #139, opened 2023-12-07
Configure test suite for edge-endpoint support

tyler-romero:


Makefile
```diff
@@ -30,20 +30,29 @@ PYTEST=poetry run pytest -v
...
 
 test-local: install  ## Run tests against a localhost API (needs GROUNDLIGHT_API_TOKEN and a local API server)
-	GROUNDLIGHT_ENDPOINT="http://localhost:8000/" ${PYTEST} ${TEST_ARGS} test
+	GROUNDLIGHT_ENDPOINT="http://localhost:8000/" ${PYTEST} ${TEST_ARGS} ${CLOUD_FILTERS} test
 
 test-integ: install  ## Run tests against the integ API server (needs GROUNDLIGHT_API_TOKEN)
-	GROUNDLIGHT_ENDPOINT="https://api.integ.groundlight.ai/" ${PYTEST} ${TEST_ARGS} test
+	GROUNDLIGHT_ENDPOINT="https://api.integ.groundlight.ai/" ${PYTEST} ${TEST_ARGS} ${CLOUD_FILTERS} test
+
+test-integ-4edge: install  ## Run tests against the integ API server (needs GROUNDLIGHT_API_TOKEN)
```
blaise-muhirwa, line 48: Just as a side note, we currently don't have support for integ on the edge since [NGINX](https://github.com/groundlight/edge-endpoint/blob/9b7f393509e57dd39a24e9c50a1e7982335d7e79/configs/nginx.conf#L22) is set up to proxy every failed request to prod. Nothing is stopping us from changing this though.

Makefile
```diff
@@ -30,20 +30,29 @@ PYTEST=poetry run pytest -v
...
+	GROUNDLIGHT_ENDPOINT="https://api.integ.groundlight.ai/" ${PYTEST} ${TEST_ARGS} ${EDGE_FILTERS} test
 
 test-docs: install  ## Run the example code and tests in our docs against the prod API (needs GROUNDLIGHT_API_TOKEN)
```
blaise-muhirwa, line 51: Do we need this filter for the docs?

test/integration/test_groundlight.py
```diff
@@ -252,6 +252,7 @@ def test_submit_image_query_with_human_review_param(gl: Groundlight, detector: D
         assert is_valid_display_result(_image_query.result)
 
 
+@pytest.mark.skip_for_edge_endpoint(reason="The edge-endpoint does not support passing metadata.")
```
blaise-muhirwa, line 255: As we were mentioned yesterday I think, we can also easily support this once we have persistent storage for image queries on the edge.

blaise-muhirwa (commented):
Looks good! Should we remove the filters for the docs make targets?

test/integration/test_groundlight.py
tyler-romero: Yeah but for now we want to unblock other PRs on the edge. And in general we may not be fully feature complete on the edge relative to the cloud.

blaise-muhirwa (approved):

PR #142, opened 2023-12-13
Removed `/python-sdk/` as the docs' base URL 

blaise-muhirwa:
This is in part because we now have more than just SDK-related content on `code.groundlight.ai`. In addition, this also allows us to just have the blog under `code.groundlight.ai/blog` instead of `code.groundlight.ai/python-sdk/blog`.


tyler-romero:
I think you missed a few references to `python-sdk`, for example this link will be broken: https://github.com/groundlight/python-sdk/blob/fc1bc450b43ea96bb0b1de959d2be054e1e951e4/docs/docs/building-applications/6-async-queries.md?plain=1#L10

Do a ctrl-f and make sure you get all of them since we're changing the base url

PR #148, opened 2024-01-06
Add detector metadata

blaise-muhirwa:
Add a field to the detector pydantic model.


src/groundlight/encodings.py
```diff
@@ -29,12 +28,12 @@ def url_encode_dict(maybe_dict: Union[Dict, str], name: str, size_limit_bytes: O
...
 
     if not isinstance(maybe_dict, dict):
-        raise TypeError(f"`{name}` must be a dictionary or JSON string: got {original_type}")
+        raise TypeError(f"`{name}` must be a dictionary or JSON string: gottt {original_type}")
 
     data_json = json.dumps(maybe_dict)
 
     if size_limit_bytes is not None:
-        size_bytes = sys.getsizeof(data_json)
+        size_bytes = len(data_json)
```
blaise-muhirwa, line 36: Made this change just to be consistent with the recent back-end change. `sys.getsizeof(data_json))` gets us the number of bytes in RAM, which is not what we want. We want to say the limit is 1024 bytes after base64 encoding.

generated/model.py
```diff
@@ -1,6 +1,6 @@
 # generated by datamodel-codegen:
 #   filename:  public-api.yaml
-#   timestamp: 2023-11-09T05:00:29+00:00
+#   timestamp: 2024-01-04T19:49:36+00:00
```
blaise-muhirwa, line 3: Initially I thought I would need to regenerate the SDK code, so I ran `make generate`, which is why the timestamp here is changed. Turns out this PR doesn't need to change the API spec.

blaise-muhirwa: In this case we don't actually need to add anything to `public-api.yaml`. It is sufficient to just add the metadata field to the `DetectorCreationInput` object, which is what I'm doing.

blaise-muhirwa: You can also see that the tests are passing without making any modifications to the API spec.

mjvogelsong: I think we want to change the `api-spec.yaml`, though, for consistency. I see that it works without it, but we'd rather have things consistent between code and spec.

spec/public-api.yaml
```diff
@@ -265,6 +265,12 @@ components:
           description:
             If the detector's prediction is below this confidence threshold,
             send the image query for human review.
+        metadata:
+          type: object 
+          nullable: true
+          description:
+            A dictionary of custom key/value metadata to associate with the detector
+            (limited to 1KB).
```
mjvogelsong, line 273: Just making it clear that it's a string representation of a dict, not a dict directly. (And similar below)

```suggestion
            A dictionary of custom key/value metadata to associate with the detector
            (limited to 1KB). This is encoded as a URL-safe, base64-encoded JSON string.
```

test/integration/test_groundlight.py
```diff
@@ -252,7 +275,52 @@ def test_submit_image_query_with_human_review_param(gl: Groundlight, detector: D
...
 
-@pytest.mark.skip_for_edge_endpoint(reason="The edge-endpoint does not support passing metadata.")
+@pytest.mark.skip_for_edge_endpoint(reason="The edge-endpoint does not support passing detector metadata.")
+def test_create_detector_with_metadata(gl: Groundlight):
+    name = f"Test {datetime.utcnow()}"  # Need a unique name
+    query = "Is there a dog?"
+    metadata = generate_random_dict(target_size_bytes=200)
+    detector = gl.create_detector(name=name, query=query, metadata=metadata)
+
+    print(f"Detector = {detector}")
```
mjvogelsong, line 285: We can remove the `prints`

mjvogelsong (approved):
LGTM

PR #156, opened 2024-01-19
Deduplicate API client docs

mjvogelsong:
For some reason, we had two clients listed on our API docs page... This removes one of them.

Screenshot 2024-01-18 at 4 46 02 PM


blaise-muhirwa (approved):
Thanks!

PR #164, opened 2024-02-08
Deprecate /device-api/posichecks endpoint

timmarkhuff:
**Background**
We need to deprecate the old /device-api/posichecks in favor of /device-api/v1/image-queries. Camilo has completed his changes to support this and they have been deployed to prod. 

**Changes in this PR**
* `spec/public-api.yaml` has been updated to support `inspection_id`
*  `submit_image_query_with_inspection` has been removed from `src/groundlight/internalapi.py` because `inspection_id` is now supported in the public API.

Jira: https://positronixcorp.atlassian.net/browse/PK-1783


src/groundlight/internalapi.py
```diff
@@ -248,68 +247,6 @@ def _get_detector_by_name(self, name: str) -> Detector:
             )
         return Detector.parse_obj(parsed["results"][0])
 
-    @RequestsRetryDecorator()
-    def submit_image_query_with_inspection(  # noqa: PLR0913 # pylint: disable=too-many-arguments
```
mjvogelsong, line 252: Yay! I like when we get to delete code!

mjvogelsong (approved):
Great, this is cleaner!

Also, I would recommend bumping the `pyproject.toml` version to `0.13.4`.

package.json
```diff
@@ -1,6 +1,8 @@
 {
   "name": "groundlight-sdk-generator",
   "dependencies": {
-    "@openapitools/openapi-generator-cli": "^2.9.0"
+    "@openapitools/openapi-generator-cli": "^2.7.0",
+    "rehype-katex": "^7.0.0",
+    "remark-math": "^6.0.0"
```
mjvogelsong, line 6: Oh, also, why were these added? Seems irrelevant to this PR.

blaise-muhirwa (approved):
Looks good to me!

robotrapta:
I think this should get lumped into a v0.14.0 release - it's a relatively big change.

PR #168, opened 2024-02-14
Faster HTTP timeout

### LINKED ISSUE #167: Absurdly long timeout if host unreachable
Here's a repro:

```
sudo echo "99.99.99.99  api.groundlightai" >> /etc/hosts
time groundlight list-detectors
```

On my ubuntu 22 box with a gigabit fiber link, it take over 8 minutes to fail!

robotrapta:
Fixes #167  - it could take over 8 minutes for an API connection to timeout if nothing's listening.


brandon-groundlight (approved):
LGTM. it feels like it would be nice to somehow change the default _request_timeout in the detectors_api and image_queries_api so we don't have to invoke the timeout parameter inside client.py, but thats more aesthetics than function

robotrapta:
I thought about exposing an environment variable to set the default timeout.  But it adds complexity, and I think it would be used extremely rarely.  The standard TCP/IP timeout values were set in the 1980s or so, when networks were extremely slow and unreliable, so it made sense to keep trying for 30 seconds to reach a remote server.  Today, that's insane, but many systems still use those classic timeout values.  Even at 5 seconds, the only place I can imagine a connection succeeding after longer than that would be on an airplane with sketchy connectivity, for example.  It's conceivable, but crazy rare.  So I don't think it's worth the complexity to even expose it.

PR #173, opened 2024-02-15
Adding python 3.12 support

robotrapta:
Also making the CI/CD tests more efficient and easier to use.  And bumping up some more versions to make other stuff more modern.

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -108,7 +108,10 @@ jobs:
   # Run integration tests against the API (only on the main branch, though). The comprehensive
   # version runs a matrix of python versions for better coverage.
   test-comprehensive:
-    if: github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v')
+    if: >-
+      github.ref == 'refs/heads/main' ||   # merges to main
+      startsWith(github.ref, 'refs/tags/v') ||   # releases
+      endsWith(github.head_ref, '-full')  # branch names requesting full testing.
     needs:
       - test-simple
       - test-docs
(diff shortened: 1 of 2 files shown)


mjvogelsong (approved):

.github/workflows/cicd.yaml
```diff
@@ -130,11 +134,12 @@ jobs:
...
             "3.10",
             "3.11",
+            "3.12",
           ]
         install_extras: [true, false]
     steps:
       - name: get code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: install python
```
mjvogelsong, line 143: I think the reason you're seeing the errors is that you're trying to install poetry bootstrapped off EOL python 3.7. You can install poetry with a modern python version, and then choose the matrix python version for the specific tested environment. Then, you should be able to use `env.POETRY_VERSION = 1.7.1`.

```yaml
      - name: install python for poetry
        uses: actions/setup-python@v4
        with:
          # We use this python to install poetry -- we choose a modern version
          python-version: ${{ env.PYTHON_VERSION }}
      - name: install poetry
        uses: snok/install-poetry@v1
        with:
          version: ${{ env.POETRY_VERSION }}
      - name: install python for python-sdk venv
        uses: actions/setup-python@v4
        with:
          # We use this python for the sdk venv
          python-version: ${{ env.PYTHON_VERSION }}
      - name: choose python version ${{ matrix.python-version }}
        run: |
          # set the environment to use the matrix python version
          poetry env use python${{ matrix.python-version }}
          poetry run python --version
```

robotrapta: Thanks Michael - that worked!

Issue #64, opened 2023-06-01
Feature request: Whoami?


robotrapta:
It would be nice to have an SDK/API method which just confirms that the API_TOKEN and ENDPOINT are configured properly.

PR #176, opened 2024-02-17
Adds whoami function

brandon-groundlight:
Addresses Feature request: Whoami? #64


generated/docs/UserApi.md
robotrapta, line 1: Is this all auto-generated?

robotrapta (approved):
This is great!  Is this just using the current `/me` endpoint used by reef?  And this is almost all generated code?

Also we should use this call in `Groundlight._verify_connectivity` instead of `list_detectors` and then we can remember who the logged in user is, and include that in the `__repr__` string so if you say 

```
gl = groundlight.Groundlight() 
print(gl)
```

it prints out the currently logged in username.  Doesn't need to be in this PR.

generated/docs/UserApi.md
brandon-groundlight: Yeah, it's mostly generated! It's neat how powerful the openapi standard is, and since we have the /me endpoint for reef this turns out to be super straightforward.

I like the rec on `_verify_connectivity` and `__repr__`, I'll add those in the next PR

PR #178, opened 2024-03-05
If our client raises a special error, we should allow users to catch it

brandon-groundlight:


blaise-muhirwa (approved):
LGTM!

blaise-muhirwa (commented):
I think you should also remove this line, which was preventing sphinx from documenting `ApiTokenError`. Now that we're exposing it, we can let sphix document it. 

https://github.com/groundlight/python-sdk/blob/687fc6fa0f344dabe2a78bea9149745fbb2f573c/sphinx_docs/models.rst?plain=1#L7

PR #179, opened 2024-03-06
Notifications in SDK

brandon-groundlight:
Part 1: Adds a couple uses of whoami that robotrapta suggested in my last PR
Part 2: Hits a new endpoint in the service to provide access to actions from the SDK. Specifically, the new function UnstableApi.create_action will be able to create new actions and configure them appropriately. I create a new API class to try and be explicit about how we anticipate this to change.


brandon-groundlight:
I think this is worth adding at this point to allow for easy programmatic access to notifications. It's not clear where this belongs though so I've created a new file for it. I intentionally didn't add it to internalapi.py because everything in internalapi.py is wired into client.py making it effectively public. Going forward there may be other functionality we want to add to the sdk and test out first and that can go into the new class as well.
The other alternative is to make an edge branch of the sdk, though I worry a branch like that could quickly diverge from main in difficult to resolve ways.

src/groundlight/unstableapi.py
```diff
@@ -0,0 +1,79 @@
...
+    Verb,
+)
+from openapi_client.model.action import Action
+from openapi_client.model.condition import Condition
+from openapi_client.model.rule_creation_input import RuleCreationInput
+
+from groundlight import Groundlight
+
+
+class UnstableApi(Groundlight):
```
brandon-groundlight, line 24: Written like this, you can call
`from groundlight import UnstableApi
gl = UnstableApi`
and gl will behave as if it was the Groundlight client but with this extra feature. The previous commit had this written into the Groundlight class but we want to make this readily apparent that this feature is unstable

src/groundlight/unstableapi.py
```diff
@@ -0,0 +1,79 @@
...
+        self,
+        detector: Union[str, Detector],
+        rule_name: str,
+        channel: Union[str, Channel],
+        recipient: str,
+        *,
+        alert_on: Union[str, Verb] = "CHANGED_TO",
+        enabled: bool = True,
+        include_image: bool = False,
+        condition_parameters: Union[str, dict] = {"label": "NO", "num_consecutive_labels": "2"},
```
brandon-groundlight, line 38: Each condition type has it's own set of condition parameters that it needs to have correct here. So if you don't know the right parameters, you'll have some trouble. The parameters match what you see from the reef webpage. The proper way to handle this would be to model each of the condition types in the openapi spec, but that's a chunk of work for something that is sounds like we might end up changing

spec/public-api.yaml
```diff
@@ -22,6 +22,41 @@ paths:
...
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/RuleCreationInput"
+          application/x-www-form-urlencoded:
+            schema:
+              $ref: "#/components/schemas/RuleCreationInput"
+          multipart/form-data:
+            schema:
+              $ref: "#/components/schemas/RuleCreationInput"
```
mjvogelsong, line 49: I think we only support `application/json`.

src/groundlight/unstableapi.py
```diff
@@ -0,0 +1,79 @@
+"""
+unstableapi.py
```
mjvogelsong, line 2: I recommend changing the wording here and elsewhere to "experimental" instead of "unstable."

src/groundlight/unstableapi.py
mjvogelsong: That's okay for now. Although, why is it a union of `str | dict` -- isn't it always a dict?

Also, using `dict` defaults leads to tricky bugs (because it's mutable). Instead, we should default to `None`, and fill the default dict inside the function.

src/groundlight/unstableapi.py
```diff
@@ -0,0 +1,79 @@
...
+            detector_id=det_id,
+            name=rule_name,
+            enabled=enabled,
+            action=action,
+            condition=condition,
+            snooze_time_enabled=snooze_time_enabled,
+            snooze_time_value=snooze_time_value,
+            snooze_time_unit=snooze_time_unit,
+        )
+        self.rules_api.create_rule(det_id, rule_input)
```
mjvogelsong, line 79:
```suggestion
        return self.rules_api.create_rule(det_id, rule_input)
```

mjvogelsong (approved):
I like the approach of using an experimental client!

spec/public-api.yaml
brandon-groundlight: This makes things easier

src/groundlight/unstableapi.py
brandon-groundlight: It's a string if someone is crazy enough to use this from the command line

As for the other point, my favorite python bug rears its head again, thanks for catching this!

PR #180, opened 2024-03-09
Expanding notifications

brandon-groundlight:
More powerful sdk access to notifications


test/unit/test_actions.py
```diff
@@ -0,0 +1,28 @@
...
+        self.assertRaises(Exception, self.gl.get_action, rule.id)
+
+    def test_get_all_rules(self):
+        det = self.gl.get_or_create_detector("test_detector", "test_query")
+        self.gl.delete_all_rules()
+        for i in range(100):
+            _ = self.gl.create_action(det, f"test_rule_{i}", "EMAIL", "test@example.com")
+        rules = self.gl.get_rules_list()
+        self.assertEqual(len(rules), 100)
+        self.gl.delete_all_rules()
```
mjvogelsong, line 26: The one issue with this is that cleanup doesn't happen if the assert fails. One cool way to make sure things are cleaned up is with `yield` in a fixture:

```python
@pytest.fixture(name="many_rules")
def _many_rules(gl: ExperimentalApi) -> Generator[List[Rule], None, None]:
    # set up the objects
    for i in range(3):
        _ = gl.create_action(det, f"test_rule_{i}", "EMAIL", "test@example.com")

    # yield the rules list for usage in a test
    yield gl.get_rules_list()

    # cleanup after the fixture is used
    gl.delete_all_rules()
```

mjvogelsong:
Oh, and we should probably fix the test failures 😂

brandon-groundlight:
merged into #185

PR #184, opened 2024-03-14
Restoring timeout functionality

timmarkhuff:
In https://github.com/groundlight/python-sdk/pull/168, Leo added faster HTTP timeouts, which was very helpful because the defaults are way too long.

In https://github.com/groundlight/python-sdk/pull/164, I was overzealous in my code deletion and accidentally removed it. This PR restores it.

Also, I found a breaking change between 0.14.0 and the RPC server/URCap. When there is no active inspection on the URCap, it sends an empty string for the inspection_id. URScript does not have the notion of None or null, or even optional arguments, so this seems to be the best we can do. Previously, the API would accept an empty string for the inspection_id, but the new API endpoint returns the following response: HTTP response body: {"inspection_id":["This field may not be blank."]}

This has been tested on the URCap/RPC server and works as expected.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ packages = [
     {include = "**/*.py", from = "src"},
 ]
 readme = "README.md"
-version = "0.14.0"
+version = "0.14.1"
 
 [tool.poetry.dependencies]
 # For certifi, use ">=" instead of "^" since it upgrades its "major version" every year, not really following semver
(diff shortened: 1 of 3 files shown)


robotrapta (approved):
While you're in here, could you change the DEFAULT_REQUEST_TIMEOUT to `10`s?  We saw some failures at 5s where it would have finished and succeeded given a bit longer.

timmarkhuff:
> While you're in here, could you change the DEFAULT_REQUEST_TIMEOUT to `10`s? We saw some failures at 5s where it would have finished and succeeded given a bit longer.

Done.

PR #185, opened 2024-03-15
V0.15.0

brandon-groundlight:
Combines some other PRs and allows:
- Full creation, deletion, and fetching of actions
- The ability to download images given their image query, solving issue #75 
- The ability to get notes and add notes to a detector

Further SDK development should NOT follow this pattern. Bouncing back and forth between the two repos is clunky to test, but more importantly we probably want to move away from manually writing out the spec.


brandon-groundlight:
tests won't pass until the appropriate backend change

test/unit/test_actions.py
```diff
@@ -0,0 +1,34 @@
...
+from groundlight import ExperimentalApi
+from openapi_client.exceptions import NotFoundException
+
+
+@pytest.fixture(name="gl")
+def _gl() -> ExperimentalApi:
+    return ExperimentalApi()
+
+
+def test_create_action(gl: ExperimentalApi):
```
mjvogelsong, line 11:
```suggestion
def test_create_rule(gl: ExperimentalApi):
```

mjvogelsong (commented):
Looks good!

The main thing is that I don't think we should change the supported python versions to 3.10+ in this PR (it will break our github action `test-comprehensive` test matrix for 3.7-3.9)

src/groundlight/experimentalapi.py
```diff
@@ -84,4 +87,82 @@ def create_action(  # pylint: disable=too-many-locals
...
+    def list_rules(self, page=1, page_size=10) -> PaginatedRuleList:
+        """
+        Gets a list of all rules
+
+        :return: a list of all rules
+        """
+        obj = self.rules_api.list_rules(page=page, page_size=page_size)
+        return PaginatedRuleList.parse_obj(obj.to_dict())
+
+    def delete_all_rules(self, detector: Union[None, str, Detector] = None) -> None:
```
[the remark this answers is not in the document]
brandon-groundlight: I think we have a request sitting around for bulk iq retrieval as well. As time permits, I'll look at adding that endpoint

pyproject.toml
```diff
@@ -9,19 +9,20 @@ packages = [
...
 frozendict = "^2.3.2"
 pillow = ">=9.0.0" # TODO: We may want to mark pillow (and numpy) as extra (https://python-poetry.org/docs/master/pyproject#extras)
 pydantic = ">=1.7.4,<3.0.0"
-python = ">=3.7.0,<4.0"
+python = ">=3.10.0,<4.0"
 python-dateutil = "^2.8.2"
 requests = "^2.28.2"
 urllib3 = "^1.26.9"
 typer = "^0.9.0"
+ipython = "^8.22.2"
```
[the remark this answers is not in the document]
brandon-groundlight: thanks for catching that, this wasn't supposed to make it in. I wanted IPython for a bit of debugging at one point. I like the idea of adding it to dev dependencies.

brandon-groundlight:
Sorry for the extra emails, the github runner was behaving differently than my local machine and took a sec to debug. But this is ready now!

src/groundlight/experimental_api.py
```diff
@@ -0,0 +1,173 @@
+"""
+experimental_t api.py
```
mjvogelsong, line 2:
```suggestion
experimental_api.py
```

src/groundlight/experimental_api.py
```diff
@@ -0,0 +1,173 @@
...
+        alert_on: Union[str, Verb] = "CHANGED_TO",
+        enabled: bool = True,
+        include_image: bool = False,
+        condition_parameters: Union[str, dict, None] = None,
+        snooze_time_enabled: bool = False,
+        snooze_time_value: int = 3600,
+        snooze_time_unit: str = "SECONDS",
+    ) -> Rule:
+        """
+        Adds a notification action to the given detector
```
mjvogelsong, line 48:
```suggestion
        Adds a notification rule to the given detector
```

test/unit/test_notes.py
```diff
@@ -0,0 +1,31 @@
+from datetime import datetime
+
+import pytest
+from groundlight import ExperimentalApi
+from openapi_client.exceptions import ApiAttributeError
+
+
+@pytest.fixture(name="gl")
+def _gl() -> ExperimentalApi:
+    return ExperimentalApi()
```
mjvogelsong, line 10: We can pull this into a common module.

test/unit/test_actions.py
```diff
@@ -0,0 +1,34 @@
...
+    rule = gl.create_rule(det, "test_rule", "EMAIL", "test@example.com")
+    rule2 = gl.get_rule(rule.id)
+    assert rule == rule2
+    gl.delete_rule(rule.id)
+    with pytest.raises(NotFoundException) as _:
+        gl.get_rule(rule.id)
+
+
+def test_get_all_actions(gl: ExperimentalApi):
+    num_test_rules = 103  # needs to be larger than the default page size
```
mjvogelsong, line 22: I don't think we should be creating this many rules in unit tests -- instead, we should patch the default page size to something smaller.

test/unit/test_notes.py
```diff
@@ -0,0 +1,31 @@
...
+            if notes[i].content == "test_note":
+                found_note = True
+        assert found_note
+    except (AssertionError, ApiAttributeError):
+        notes = gl.get_notes(det)["customer"]
+        found_note = False
+        for i in range(len(notes)):
+            if notes[i].content == "test_note":
+                found_note = True
+        assert found_note
```
mjvogelsong, line 31: Does something like this work?

```suggestion
    notes = (gl.get_notes(det).get("gl") or []) + (gl.get_notes(det).get("customer") or [])
    found_note = False
    for i in range(len(notes)):
        if notes[i].content == "test_note":
            found_note = True
    assert found_note
```

mjvogelsong (approved):

test/unit/test_notes.py
brandon-groundlight: ooh, that's nice

brandon-groundlight:
YEEEAAAAAH!

PR #196, opened 2024-03-28
Thought dev dependencies should be ignored

brandon-groundlight:
I thought poetry would ignore ipython in the dev dependency unless specifically asked. This fixes the toml and regenerates the lockd


paulina-positronix (approved):
Thank you!

PR #197, opened 2024-03-28
Unique detector name

brandon-groundlight:
Make detector names unique so that the test matrix doesn't reach for the same test at the same time


blaise-muhirwa (approved):
LGTM!

brandon-groundlight:
Turns out this doesn't fix the problem, but it's good code so I'll merge it anyways

PR #199, opened 2024-04-02
makes test idempotent

brandon-groundlight:
If the test failed partway through, it may leave rules lying around which affect the next test run


test/unit/test_actions.py
```diff
@@ -4,6 +4,7 @@
 
 
 def test_create_action(gl: ExperimentalApi):
+    gl.delete_all_rules()
```
blaise-muhirwa, line 7: Can we add a comment here mentioning that we are deleting rules first just so previous runs don't affect results from the current test run?

blaise-muhirwa (approved):
LGTM!

PR #201, opened 2024-04-04
Rename groundlight_openapi_client

mjvogelsong:
OpenAPI generator (annoyingly) uses `openapi_client` as a top-level import, which forces us to expose `openapi_client` as a top-level package when it's pip installed. This causes issues if a consumer also uses `openapi_client`.

So, as a workaround, we can rename `openapi_client` to `groundlight_openapi_client` to prevent name clashes. There might be a better way to do this, but we're not sure how yet.


Makefile
```diff
@@ -19,7 +19,8 @@ install-generator: install ## Install dependencies for SDK code generator
 generate: install-generator  ## Generate the SDK from our public openapi spec
 	node_modules/.bin/openapi-generator-cli generate -i spec/public-api.yaml \
 		-g python \
-		-o ./generated
+		-o ./generated \
+		--additional-properties=packageName=groundlight_openapi_client
```
mjvogelsong, line 23: This is the main change. Everything else is either auto-generated or CTRL-F changing references from `openapi_client` to `groundlight_openapi_client`

brandon-groundlight (approved):
Nice, this is has to be the record for number of files changed. I love autogenerated code

PR #203, opened 2024-04-09
[PK-3193] - Fix redirect with the Groundlight logo and change the script for Google Analytics

jesusvalle-lw:
This change is to redirect to the index page instead of opening a new tab when the user clicks on the Groundlight logo.


Dexcrash (approved):

JhoamMolina (approved):
LGTM!

PR #209, opened 2024-05-10
Bump some dep versions

tyler-romero:
Bumping some dependency versions. Specifically the old version of typer is conflicting with something I need. And only the latest version of pydantic V1 (1.10) is getting security updates still.


test/unit/test_cli.py
```diff
@@ -7,7 +7,7 @@
 
 def test_whoami():
     completed_process = subprocess.run(
-        ["groundlight", "whoami"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True
+        ["groundlight", "whoami"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True, check=False
```
tyler-romero, line 10: I think these are all b/c I bumped the ruff version

brandon-groundlight (approved):
No issue from me

PR #213, opened 2024-06-01
Api spec rework

brandon-groundlight:
We recently did a full re-haul of the public api spec. Functionality of the SDK should be unaffected, but this maps better onto what the service provides and sets up for improvements in the future.


Makefile
```diff
@@ -21,7 +21,7 @@ generate: install-generator  ## Generate the SDK from our public openapi spec
 		-g python \
 		-o ./generated \
 		--additional-properties=packageName=groundlight_openapi_client
-	poetry run datamodel-codegen  --input spec/public-api.yaml --output generated/model.py
+	poetry run datamodel-codegen  --input spec/public-api.yaml --output generated/model.py --strict-nullable
```
tyler-romero, line 24: What does --strict-nullable do? For less obvious flags its helpful to comment them or link to their documentation

spec/public-api.yaml
```diff
@@ -1,283 +1,243 @@
...
-      security:
-        - ApiToken: []
-      responses:
-        "204":
-          description: ""
-  /v1/actions/detector/{detector_id}/rules:
+                $ref: '#/components/schemas/PaginatedRuleList'
+          description: ''
     post:
       operationId: Create rule
```
tyler-romero, line 34:
```suggestion
      operationId: Create detector rule
```

spec/public-api.yaml
```diff
@@ -1,283 +1,243 @@
...
-                $ref: "#/components/schemas/Rule"
-          description: ""
-  /v1/notes:
+                $ref: '#/components/schemas/Rule'
+          description: ''
+  /v1/actions/rules:
     get:
-      operationId: get notes
-      description: Retrieve notes for a detector
+      operationId: List rules
```
tyler-romero, line 67: Which rules? All the rules for a user? Or a group?

src/groundlight/client.py
```diff
@@ -266,13 +266,15 @@ def create_detector(  # noqa: PLR0913
...
-        if pipeline_config is not None:
-            detector_creation_input.pipeline_config = pipeline_config
+        detector_creation_input = DetectorCreationInputRequest(
+            name=name,
+            query=query,
+            pipeline_config=pipeline_config,
+        )
         if metadata is not None:
             detector_creation_input.metadata = str(url_encode_dict(metadata, name="metadata", size_limit_bytes=1024))
+        if confidence_threshold:
```
tyler-romero, line 276: Was there a reason to reorder this? Not saying you need to change it just wondering if Im missing something

src/groundlight/experimental_api.py
```diff
@@ -9,23 +9,26 @@
...
-from groundlight_openapi_client.model.note_creation_input import NoteCreationInput
-from groundlight_openapi_client.model.rule_creation_input import RuleCreationInput
-from model import Channel, Detector, PaginatedRuleList, Rule, Verb
+from groundlight_openapi_client.model.action_request import ActionRequest
+from groundlight_openapi_client.model.channel_enum import ChannelEnum
+from groundlight_openapi_client.model.condition_request import ConditionRequest
+from groundlight_openapi_client.model.note_request import NoteRequest
+from groundlight_openapi_client.model.rule_request import RuleRequest
+from groundlight_openapi_client.model.verb_enum import VerbEnum
+from model import Detector, PaginatedRuleList, Rule
```
tyler-romero, line 22: Any idea why this is still considered Experimental? Should we change that?

spec/public-api.yaml
tyler-romero: Just trying to add some parallelism to the name/description patterns

Makefile
brandon-groundlight: Yeah, let me add a comment in the file itself, but the useful discussion is [here](https://github.com/koxudaxi/datamodel-code-generator/issues/327). For properties in the api spec that have `nullable: true` the generated pydantic class should let that property be Optional. For strange reasons, you need to set --strict-nullable for this to be true.

spec/public-api.yaml
brandon-groundlight: Users and groups should share the exact same set of rules, so I'm not sure if that should be differentiated

src/groundlight/client.py
brandon-groundlight: No, I think I just happened to move it down while I was moving the pipeline_config default into the constructor

src/groundlight/experimental_api.py
brandon-groundlight: At some point this moves into the main client body, but I think we just haven't had the impetus to change this yet. This was our way of leaving a door open in case we wanted to change things later

brandon-groundlight:
> Its not self explanatory what this improvement is setting up for

This is mainly the corresponding SDK update to the new process of generating the api spec. The generated code had been rotting for a while, so this is mainly cleaning that before we attach more onto this. The next likely change (and slightly overdue) change is to make confidence non-nullable.

spec/public-api.yaml
brandon-groundlight: I agreed with you at first, but then I realized this broke the parallelism to the delete rule endpoint. It's implicit that all rules belong to a detector

PR #214, opened 2024-06-04
Convenience update

brandon-groundlight:
Adds a few new convenience parameters when setting up a detector. Specifically, we want users to be able to specify the group_name and the patience time


timmarkhuff (approved):
LTGM. There was a CI failure, but once that's fixed, I would say it's good to merge.

I do worry a little bit that we raise exceptions when the group_name, confidence_threshold etc. are different than the existing detector. That sets the user up for a situation where their code works fine for a while, and then because of a change another user makes through the webapp, the code breaks. That said, I'm not sure what the right solution is. Perhaps:
1. providing a value that differs from the existing detector's value updates that value
2. don't include any of these parameters in this method. Have separate setter methods for these values (like how we already have one for setting the detector confidence)

Of the two options, I would probably vote for the first.

src/groundlight/client.py
```diff
@@ -256,8 +258,12 @@ def create_detector(  # noqa: PLR0913
 
         :param query: the detector query
 
+        :param group_name: the detector group that the new detector should belong to
```
tyler-romero, line 261: Should add what `None` defaults to

src/groundlight/client.py
```diff
@@ -256,8 +258,12 @@ def create_detector(  # noqa: PLR0913
...
+
         :param confidence_threshold: the confidence threshold
 
+        :param patience_time: the patience time, or how long Groundlight should work to generate a confident answer
```
tyler-romero, line 265: Same here, what is the default?

tyler-romero (approved):

PR #217, opened 2024-06-19
removing py3.7 support

brandon-groundlight:
Python 3.7 has been End-of-Life (EOL) for almost a year now.  It finally broke in our tests because of one of our dependent libraries, so it's time to say goodbye.


robotrapta (approved):

tyler-romero (approved):
LGTM after updating the lockfile!

robotrapta:
This is a breaking change, so we should bump version to 0.17

PR #219, opened 2024-06-25
Limit edge endpoint test

brandon-groundlight:
The edge endpoint doesn't support creating detectors, which causes a number of sdk tests to fail. However, we've successfully been using the edge endpoint, indicating that the failing tests are not actually testing what we want to test. This PR is a stopgap measure to get all tests passing so we can continue development.

In the future:
1. We need to better define which tests are run against the edge endpoint. We have a problem right now where the sdk defines the tests for the edge endpoint, but sdk development can happen independently of the edge endpoint. I think we either want to define which tests are run against the edge endpoint outside of the sdk, or force all sdk development to support the current edge-endpoint. This is quite a bit of lift though, and should be done in another PR.
2. There's not a great reason I'm aware of that the edge endpoint shouldn't be able to forward sdk requests to the main service. Some of the tests I'm blocking right now could reasonably run against the edge-endpoint, but we need to update the edge-endpoint to support these features.

### DIFF
--- a/test/integration/test_groundlight.py
+++ b/test/integration/test_groundlight.py
@@ -99,6 +99,7 @@ def fixture_image() -> str:
     return "test/assets/dog.jpeg"
 
 
+@pytest.mark.skip_for_edge_endpoint(reason="The edge-endpoint does not support creating detectors.")
 def test_create_detector(gl: Groundlight):
     name = f"Test {datetime.utcnow()}"  # Need a unique name
     query = "Is there a dog?"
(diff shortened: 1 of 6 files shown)


brandon-groundlight:
The test failures didn't run as deep as I thought, and this PR should fix the test failures in edge-endpoint

tyler-romero:
Rather than skipping tests wholesale, can we just update the edge endpoint? I feel this way especially about the `edge-endpoint` not being compatible with the latest sdk. One of the requirements for the sdk should be that it works with the `edge-endpoint` and the cloud.

brandon-groundlight:
Talked with Tyler and Harry. We'd rather leave the tests failing for the time being than skip on these tests. A more complete fix will follow

PR #220, opened 2024-07-01
Skipped Inspection Tests for Edge

honeytung:
Edge does not support inspections so this PR is skipping those tests.


brandon-groundlight (approved):
I'm a little weirded out that all of these were NOT failing before, and only 2 of them were actually failing. Regardless, since we don't formally support inspections on the edge-endpoint (and never have) it makes sense to turn off these tests for the time being. We can swing back later as the need arises.

PR #228, opened 2024-07-19
Adds the experimentalapi client class to the docs

brandon-groundlight:
This simply manually adds the experimental sections into the instructions for Sphinx. We haven't tightly integrated Sphinx yet, so in the future we should look into how to make this process more automatic


generated/model.py
```diff
@@ -81,6 +106,11 @@ class ActionRequest(BaseModel):
 
 
 class AllNotes(BaseModel):
+    """
+    Serializes all notes for a given detector, grouped by type as listed in UserProfile.NoteCategoryChoices
+    The fields must match whats in USERPROFILE.NoteCategoryChoices
```
sunildkumar, line 111: `USERPROFILE`

is all caps right?

Makefile
```diff
@@ -22,7 +22,7 @@ generate: install-generator  ## Generate the SDK from our public openapi spec
 		-o ./generated \
 		--additional-properties=packageName=groundlight_openapi_client
 # strict-nullable makes nullable fields Optional in the generated Pydantic classes: https://github.com/koxudaxi/datamodel-code-generator/issues/327
-	poetry run datamodel-codegen  --input spec/public-api.yaml --output generated/model.py --strict-nullable
+	poetry run datamodel-codegen  --input spec/public-api.yaml --output generated/model.py --strict-nullable --use-schema-description
```
sunildkumar, line 25: what does this new flag do?

src/groundlight/experimental_api.py
```diff
@@ -26,6 +26,11 @@
 
 class ExperimentalApi(Groundlight):
     def __init__(self, endpoint: Union[str, None] = None, api_token: Union[str, None] = None):
+        """
+        Constructs an experimental groundlight client. The experimental client inherits all the functionality of the
+        base groundlight client, but also includes additional functionality that is still in development. Additional
+        functionality is subject to change
+        """
```
sunildkumar, line 33:
```suggestion
        """
        Constructs an experimental groundlight client. The experimental client inherits all the functionality of the
        base groundlight client, but also includes additional functionality that is still in development. Experimental
        functionality is subject to change.
        """
```

sunildkumar (approved):
lgtm. Left a couple suggestions and a question.

Dexcrash (approved):
LGTM

Makefile
brandon-groundlight: It uses the schema description to generate docstrings on the generated classes

generated/model.py
brandon-groundlight: Well, no... In fact, SDK users don't have access to UserProfiles so we shouldn't even mention that here. But the source of truth for this docstring lives in the BE, so that will require a separate change

PR #230, opened 2024-07-31
Bbox support

brandon-groundlight:
Adds support for providing bounding boxes and reading them in image queries


brandon-groundlight:
There will be test failures until the BE change is deployed

tyler-romero (approved):
LGTM! triggerd the cicd tests again to see if they're succeeding now that the other half of this is in prod

brandon-groundlight:
We need an update before tests pass to reflect renaming in the BE

PR #232, opened 2024-08-12
Allow submiting images alongside notes in the sdk

brandon-groundlight:
We previously allowed users to create text notes for detectors. (Notes allow users to provide additional instruction for their task, which can aid human reviewers and improve ML performance). Now users can attach an image as well!


timmarkhuff (approved):
LGTM. I tried it out and it was pretty straightforward. I wonder if we want to allow users to submit images in other formats (PIL and numpy) like we do in other situations, but it's not a big deal.

src/groundlight/client.py
```diff
@@ -731,13 +731,17 @@ def add_label(
...
     ):
         """
-        Add a new label to an image query.  This answers the detector's question.
+        Add a new label to an image query.  This answers the detector's
+        question.
 
-        :param image_query: Either an ImageQuery object (returned from `submit_image_query`)
-                            or an image_query id as a string.
+        :param image_query: Either an ImageQuery object (returned from
+                            `submit_image_query`) or an image_query id as a
```
sunildkumar, line 738: > returned from `submit_image_query`

My understanding is that we encourage users to use `ask_*` methods instead?

src/groundlight/experimental_api.py
```diff
@@ -180,16 +181,26 @@ def get_notes(self, detector: Union[str, Detector]) -> Dict[str, Any]:
         det_id = detector.id if isinstance(detector, Detector) else detector
         return self.notes_api.get_notes(det_id)
 
-    def create_note(self, detector: Union[str, Detector], note: Union[str, NoteRequest]) -> None:
+    def create_note(self, detector: Union[str, Detector], note: str, image: Union[str, None] = None) -> None:
```
sunildkumar, line 184: nit: I though `Union` was replaced with `|`, e.g. `detector: str | Detector`. Maybe this doesn't work for some of the python versions we support?

PR #236, opened 2024-08-21
Skipped detector metadata testing for edge

honeytung:
This PR just add a line to skip the detector metadata test for `edge-endpoint` since it is not supported.


brandon-groundlight (approved):
Awesome, looks good

PR #241, opened 2024-08-26
Make human_review available in ask_ml

brandon-groundlight:
This is counterintuitive, having a function named ask_ml have a parameter called human_review. This might tell us we may need to review the intended mental model for sdk users


brandon-groundlight:
Chatted more with Sunil, I'll leave this PR open in case have strong reason to make this change to unblock key users. But this process feels like it adds a level of complexity that undermines the goal of having simple ask_* methods. While this change exposes new functionality to ask_ml, this doesn't add novel functionality to the sdk as equivalent behavior can be obtained with other functions in the SDK

PR #242, opened 2024-08-26
Reset detector

brandon-groundlight:
SDK support for resetting detectors


tyler-romero (approved):
LGTM, I really wish we didnt have to skip those tests though. How long does the `test_reset_retry` test take if you poll every second instead of every minute?

brandon-groundlight:
Thanks for the catch Tyler, what I initially thought was a delay was actually a BE bug that Roxanne just fixed. Turned the tests back on

PR #260, opened 2024-10-07
Makes source no longer required

brandon-groundlight:
This should in theory make v0.18 backwards. compatible


brandon-groundlight:
I need to check in on the test runner's permissions before I can merge this in, but the branch is here to unblock internal work

brandon-groundlight:
Test runners are updated, should be good to go

tyler-romero:
The fact that source is optional is very unintuitive I think (and hard to determine why by reading the code). Is it possible to add a tag or annotation or something that explians its for compatibility between x version and y version and that in fact it will be returned in y version?

tyler-romero (approved):
Discussed offline, LGTM after adding some sort of comment about why source is optional and when the version cutoff is

PR #266, opened 2024-10-15
Skips some tests that are currently causing problems for internal users

brandon-groundlight:


test/unit/test_detector_reset.py
```diff
@@ -6,6 +6,7 @@
 from groundlight_openapi_client.exceptions import NotFoundException
 
 
+@pytest.mark.skip(reason="This is an expensive test, reset may take some time")
```
brandon-groundlight, line 9: This is likely the source of much of our current SDK test time

test/unit/test_detector_reset.py
```diff
@@ -25,6 +26,7 @@ def test_reset_retry(gl_experimental: ExperimentalApi):
         raise Exception("Failed to reset detector")
 
 
+@pytest.mark.skip(reason="This test does not work with strong 0 shot models, enabled by default based on your account")
```
brandon-groundlight, line 29: This no longer works for internal users because GLAVA is so confident

test/unit/test_experimental.py
```diff
@@ -56,6 +56,12 @@ def test_counting_detector(gl_experimental: ExperimentalApi):
     assert count_iq.result.count is not None
 
 
+@pytest.mark.skip(
```
brandon-groundlight, line 59: The SDK got ahead of reef, and what we've enabled for internal users, so this test is actually causing issues during development

brandon-groundlight (commented):
SDK tests run in 3 min 22 seconds on my local machine after these changes. I've made a ticket to organize the tests better so we can call only a certain subset of tests based on who is making the call.

tyler-romero (approved):
LGTM! Thanks Brandon!

PR #267, opened 2024-10-15
Python3.12 pillow dependencies

brandon-groundlight:
I'm confused as to why this stopped working in the first place. But the package install matches the recommendation from Pillow [here](https://pillow.readthedocs.io/en/stable/installation/building-from-source.html#external-libraries)


tyler-romero (approved):

brandon-groundlight:
Also see here for the currently failing action and that it's not transient
https://github.com/groundlight/python-sdk/actions/runs/11333942780

PR #271, opened 2024-10-30
Allow specifying an ID when submitting an image query

CoreyEWood:
Adds optional parameter `image_query_id` to methods which allow submitting an image query. Adds tests for new parameter. Updates generated code based on new spec.

(won't be merged until the backend changes are in production)


Makefile
```diff
@@ -54,6 +54,9 @@ test-local: install  ## Run tests against a localhost API (needs GROUNDLIGHT_API
 test-integ: install  ## Run tests against the integ API server (needs GROUNDLIGHT_API_TOKEN)
 	GROUNDLIGHT_ENDPOINT="https://api.integ.groundlight.ai/" ${PYTEST} ${TEST_ARGS} ${CLOUD_FILTERS} test
 
+test-dev: install ## Run tests against a dev API server (needs GROUNDLIGHT_API_TOKEN and properly configured dns-hostmap)
```
CoreyEWood, line 57: Added this because it seems like it might be useful for the future, but let me know if we don't want this in here.

brandon-groundlight: I'm indifferent on this one. Only thought I have is that given the above two commands you could instead make these commands set a variable and then call `make test` using that variable

spec/public-api.yaml
```diff
@@ -995,6 +998,7 @@ components:
       - confidence
       - created_at
       - detector_id
+      - source
```
brandon-groundlight, line 1001: Just noting other conversations we've had, this is unrelated to the goal of this PR, but brings this doc back in sync with the BE

test/integration/test_groundlight.py
```diff
@@ -320,6 +320,53 @@ def test_submit_image_query_png(gl: Groundlight, detector: Detector):
...
+    # submit_image_query
+    id = f"iq_{KsuidMs()}"
+    _image_query = gl.submit_image_query(
+        detector=detector.id, image="test/assets/dog.jpeg", wait=10, human_review="NEVER", image_query_id=id
+    )
+    assert str(_image_query)
+    assert isinstance(_image_query, ImageQuery)
+    assert is_valid_display_result(_image_query.result)
+    assert _image_query.metadata is not None
+    assert _image_query.metadata.get("is_from_edge")
```
brandon-groundlight, line 333: Why not check that the returned ID is actually the one you submitted?

brandon-groundlight (commented):
Looks pretty good

Unless we want non-edge users to be specifying ids, I think we should exclude it as a parameter everywhere except for submit_image_query. The other functions are simply convenience wrappers for submit_image_query and having image query id as a parameter is a hit to their ease of use

test/integration/test_groundlight.py
CoreyEWood: Good point, it definitely should check that!

Makefile
CoreyEWood: Good idea, I made some changes accordingly - let me know if those are what you were envisioning.

brandon-groundlight (approved):
Thanks! LGTM

PR #279, opened 2024-11-30
Documenation Improvements and Updates

tyler-romero:
# How to Review:
This is a large documentation PR with lots of changes to formatting, wording, and site organization. In order to effectively review,  please check out this branch (`tyler/more-code-samples`) and run `make develop-docs-comprehensive` to see a development version of the site.

# Changes (selected)
1. Update documentation to reflect that we support Python 3.9+
2. Document usage of `ask_ml` and `ask_confident`
3. Reorganize site to improve consistency and flow (most significantly, docs for *developing* applications are separate from docs detailing *sample* applications)
4. Document use of `framegrab` and tuning of motion detection
5. Document `Alerts`, update documentation for the `edge-endpoint` and `groundlight/stream`

PR #281, opened 2024-12-02
Adding py3.13 to tests

robotrapta:
Adding python 3.13 to the test matrix.  Updating old references to versions of python <3.9 which we don't support any more.


tyler-romero:
Can you wait for https://github.com/groundlight/python-sdk/pull/279 ? This includes all the updates to 3.8 -> 3.9

robotrapta:
> Can you wait for https://github.com/groundlight/python-sdk/pull/279 ? This includes all the updates to 3.8 -> 3.9

Sure thing.  I can rebase on top of it even.

generated/setup.py
```diff
@@ -32,7 +32,7 @@
     author_email="support@groundlight.ai",
     url="",
     keywords=["OpenAPI", "OpenAPI-Generator", "Groundlight API"],
-    python_requires=">=3.6",
+    python_requires=">=3.9",
```
tyler-romero, line 35: I think this is the only change I didnt capture - although this file is generated so I'm not sure if its the right place to make the change.

robotrapta: Yeah, this file AFAIK isn't actually used by anything.

robotrapta:
Tyler beat me to it!

PR #282, opened 2024-12-03
Remove imghdr

brandon-groundlight:
Give a 21 print line salute to imghdr which is slated for removal from the standard library


tyler-romero (approved):
LGTM! thanks!

PR #284, opened 2024-12-04
Bump to version 0.20.0

tyler-romero:


brandon-groundlight (approved):
I've not figured out what the best way to do these PRs is. It feels bad to have a whole PR for this, but I always forget to bump it in the PR right before I publish a new version.

Anyways, I think about this too much. LGTM

tyler-romero:
FWIW in other libraries I see maintainers deliberately keep version bumps in their own PRs

PR #285, opened 2024-12-04
Multiclass label support

brandon-groundlight:
Tests and labeling support for multiclass detectors


src/groundlight/binary_labels.py
```diff
@@ -53,25 +53,3 @@ def convert_internal_label_to_display(
 
     logger.warning(f"Unrecognized internal label {label} - leaving it alone as a string.")
     return label
-
-
-def convert_display_label_to_internal(
```
tyler-romero, line 58: 🥳

src/groundlight/client.py
```diff
@@ -1066,16 +1068,20 @@ def _wait_for_result(
...
+        rois: Union[List[ROI], str, None] = None,
     ):
         """
         Provide a new label (annotation) for an image query. This is used to provide ground-truth labels
         for training detectors, or to correct the results of detectors.
 
         **Example usage**::
 
-            gl = Groundlight()
+            gl = ExperimentalApi()
```
tyler-romero, line 1084: Why does this need to be `ExperimentalAPI()`

tyler-romero (approved):
LGTM! Thanks for adding those new tests too

src/groundlight/client.py
brandon-groundlight: Copy paste, my greatest power and greatest weakness

tyler-romero:
I'm seeing backend errors that I believe are related to this PR. Specifically, submitting counting labels triggers count-detector training. But our create_count_detector sdk method doesnt allow specifying what the object to be counted is. This bad specification of count detectors causes errors during training.

test/unit/test_labels.py
```diff
@@ -0,0 +1,58 @@
...
+    gl_experimental.add_label(iq1, "apple")
+    iq1 = gl_experimental.get_image_query(iq1.id)
+    assert iq1.result.label == "apple"
+    gl_experimental.add_label(iq1, "banana")
+    iq1 = gl_experimental.get_image_query(iq1.id)
+    assert iq1.result.label == "banana"
+    gl_experimental.add_label(iq1, "cherry")
+    iq1 = gl_experimental.get_image_query(iq1.id)
+    assert iq1.result.label == "cherry"
+    # You can submit the index of the class as well
```
srnangi, line 53: Ohh, I see that we can use class index in labels. Should we add this to function description for `add_label` too?

srnangi (approved):
LGTM! Thanks.

PR #286, opened 2024-12-05
Include confidence_threshold in submit_image_query params

CoreyEWood:
Pass `confidence_threshold` through to the `submit_image_query` call, which will fix some edge behavior and improve overall consistency. Will merge after the backend change is deployed.


spec/public-api.yaml
```diff
@@ -1367,8 +1375,7 @@ components:
           - ALGORITHM
         count:
           type: integer
-          minimum: null
```
brandon-groundlight, line 1370: I made a fix earlier in the week, these are just plainly invalid swagger that is corrected

brandon-groundlight (approved):
LGTM

spec/public-api.yaml
brandon-groundlight: Incidentally, to whom it may concern, this is a great website
https://editor.swagger.io/

PR #287, opened 2024-12-05
Remove unused ci files

brandon-groundlight:
dynamic updating these files involves editing the mustache templates, a hill we're not going to charge at this time


CoreyEWood (approved):
Seems good to me 👍

PR #293, opened 2024-12-09
Update docstring

brandon-groundlight:


src/groundlight/client.py
```diff
@@ -743,7 +743,7 @@ def ask_confident(  # noqa: PLR0913 # pylint: disable=too-many-arguments
           converted to JPEG at high quality before sending to service.
         :param confidence_threshold: The confidence threshold to wait for.
             If not set, use the detector's confidence threshold.
-        :param wait: How long to wait (in seconds) for a confident answer.
+        :param wait: How long to wait (in seconds) for a confident answer. Default is 30.0.
```
roxanne-o, line 746: This is confusing because the actual default value is None. So maybe you should say explicitly that if it's None, we default to 30. A better solution would be the default value being 30.

PR #296, opened 2024-12-12
Internal function bugfix

brandon-groundlight:
Considering a case where ask_async returned image queries have no result


tyler-romero (approved):
LGTM!

PR #299, opened 2024-12-19
quick fix to support counting unclear in the SDK

brandon-groundlight:
This is a potential patch for handling unclear results for counting detectors


tyler-romero (approved):
LGTM, thanks for the fix.

PR #302, opened 2025-01-10
Create multiple actions on alert

brandon-groundlight:
A followup for later, we will completely replace 'rule' with 'alert'. Once renamed we can move alert functions into the standard client. This requires a bit of backend modelling updates.

I'm not a huge fan of root as the attribute of ActionList, but that matches the style of datamodel-codegen.

### DIFF
--- a/src/groundlight/experimental_api.py
+++ b/src/groundlight/experimental_api.py
@@ -16,6 +16,7 @@ from groundlight_openapi_client.api.detector_groups_api import DetectorGroupsApi
 from groundlight_openapi_client.api.detector_reset_api import DetectorResetApi
 from groundlight_openapi_client.api.image_queries_api import ImageQueriesApi
 from groundlight_openapi_client.api.notes_api import NotesApi
+
 from groundlight_openapi_client.model.action_request import ActionRequest
 from groundlight_openapi_client.model.channel_enum import ChannelEnum
 from groundlight_openapi_client.model.condition_request import ConditionRequest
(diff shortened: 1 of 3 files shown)


test/integration/test_groundlight.py
```diff
@@ -37,7 +37,8 @@ def is_valid_display_result(result: Any) -> bool:
         and not isinstance(result, MultiClassificationResult)
     ):
         return False
-    if not is_valid_display_label(result.label):
```
brandon-groundlight, line 40: separate change that needed to get in to support different result types

src/groundlight/experimental_api.py
```diff
@@ -93,6 +104,142 @@ def __init__(
...
+        human_review_required: bool = False,
+    ) -> Rule:
+        """
+        Creates an alert for a detector that will trigger actions based on specified conditions.
+
+        An alert allows you to configure automated actions when certain conditions are met,
+        such as when a detector's prediction changes or maintains a particular state.
+
+        .. note::
+            Currently, only binary mode detectors (YES/NO answers) are supported for notification rules.
```
f-wright, line 179:
```suggestion
            Currently, only binary mode detectors (YES/NO answers) are supported for alerts.
```

src/groundlight/experimental_api.py
```diff
@@ -93,6 +104,142 @@ def __init__(
...
+            )
+            alert = gl.create_alert(
+                detector="det_idhere",
+                name="Door Open Alert",
+                condition=condition,
+                actions=[action1, action2]
+            )
+
+        :param detector: The detector ID or Detector object to add the rule to
+        :param name: A unique name to identify this rule
```
f-wright, line 208:
```suggestion
        :param name: A unique name to identify this alert
```

src/groundlight/experimental_api.py
```diff
@@ -93,6 +104,142 @@ def __init__(
...
+        """
+        Creates an Action object for use in creating alerts
+
+        This function serves as a convenience method; Action objects can also be created directly.
+
+        **Example usage**::
+
+            gl = ExperimentalApi()
+
+            # Create an action for a rule
```
f-wright, line 146:
```suggestion
            # Create an action
```

f-wright (approved):
LGTM! I left a few suggestions to remove references to rules in the new create_alert function. If we're switching to use exclusively "alert", that seems like it could become confusing.

PR #303, opened 2025-01-13
tells users about GROUNDLIGHT_ENDPOINT in client docstring

brandon-groundlight:


timmarkhuff (approved):
LGTM

I wonder if we should be more generic than this. Couldn't someone also use a different endpoint if they were using private cloud, not just edge? I'm probably overthinking this...

PR #309, opened 2025-01-18
Webhooks for alerts

f-wright:
Webhooks have been recently added as an action option for Groundlight, this PR makes them accessible through the SDK.


src/groundlight/experimental_api.py
```diff
@@ -46,7 +48,7 @@
 from .client import DEFAULT_REQUEST_TIMEOUT, Groundlight, logger
 
 
-class ExperimentalApi(Groundlight):
+class ExperimentalApi(Groundlight):  # pylint: disable=too-many-public-methods
```
f-wright, line 51: Let me know if there's a different way you'd prefer I solve this -- it seems best to keep the method here so I just disabled it.

brandon-groundlight: I think I prefer this. The pylint suggestion is probably to group up the methods under subclasses, but I think that's unnecessary at this time

src/groundlight/experimental_api.py
```diff
@@ -156,12 +158,29 @@ def make_action(
...
+        )
+
+    def create_alert(  # pylint: disable=too-many-locals, too-many-arguments  # noqa: PLR0913
         self,
         detector: Union[str, Detector],
         name,
         condition: Condition,
-        actions: Union[Action, List[Action], ActionList],
+        actions: Optional[Union[Action, List[Action], ActionList]] = None,
+        webhook_actions: Optional[Union[WebhookAction, List[WebhookAction]]] = None,
```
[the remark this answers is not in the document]
f-wright: Yeah, I agree with all of that. It is confusing for the user, but it seems like this makes the most sense in terms of keeping the SDK in line with the backend infra. 

If we do want to decouple the SDK from the rest of the infra in this way, I think it could be fine to create a convenience method for the users that allows them to submit a single list of actions, and we can split them into separate "actions" and "webhook actions" before sending to the back end.

PR #313, opened 2025-02-08
Allows binary download for edge

brandon-groundlight:


generated/test/test_edge_model_info.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def setUp(self):
+        pass
+
+    def tearDown(self):
+        pass
+
+    def testEdgeModelInfo(self):
+        """Test EdgeModelInfo"""
+        # FIXME: construct object with mandatory attributes with example values
```
paulina-positronix, line 29: Is this supposed to be replaced with an actual test?

generated/test/test_edge_api.py
```diff
@@ -0,0 +1,32 @@
...
+    """EdgeApi unit test stubs"""
+
+    def setUp(self):
+        self.api = EdgeApi()  # noqa: E501
+
+    def tearDown(self):
+        pass
+
+    def test_get_model_urls(self):
+        """Test case for get_model_urls"""
```
paulina-positronix, line 27: Is this supposed to be replaced with an actual test?

generated/test/test_model_url.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def setUp(self):
+        pass
+
+    def tearDown(self):
+        pass
+
+    def testModelUrl(self):
+        """Test ModelUrl"""
+        # FIXME: construct object with mandatory attributes with example values
```
paulina-positronix, line 29: Is this supposed to be replaced with an actual test?

src/groundlight/experimental_api.py
```diff
@@ -947,3 +952,48 @@ def create_multiclass_detector(  # noqa: PLR0913 # pylint: disable=too-many-argu
         detector_creation_input.mode_configuration = mode_config
         obj = self.detectors_api.create_detector(detector_creation_input, _request_timeout=DEFAULT_REQUEST_TIMEOUT)
         return Detector.parse_obj(obj.to_dict())
+
+    def _download_mlbinary_url(self, detector: Union[str, Detector]) -> EdgeModelInfo:
+        """
+        Gets a temporary presigned URL to download the model binaries for the given detector, along
+        with relevant metadata
```
paulina-positronix, line 959: Do you need secrets or anything to do this? Or is that taken care of here?

paulina-positronix (commented):
It seems like the main change is inside experimental_api.py which LGTM. I'm not sure about the rest of the generated API code and especially the empty tests.

src/groundlight/experimental_api.py
brandon-groundlight: The secrets are taken care of in the backend after you've authenticated with your api token. So long as you have the feature enabled in your Groundlight account, you should be able to download

brandon-groundlight:
> It seems like the main change is inside experimental_api.py which LGTM. I'm not sure about the rest of the generated API code and especially the empty tests.

True that the generated tests aren't particularly useful. Nothing inside the generated folder is written by hand, and in fact editing those files manually is very likely to get those changes overwritten, which is why we don't update the tests there.

paulina-positronix (approved):

PR #316, opened 2025-02-21
Doc fix for create detector

brandon-groundlight:
Quick fix to make the other create detector methods more easily discovered


src/groundlight/client.py
```diff
@@ -207,6 +207,7 @@ def _verify_connectivity(self) -> None:
                 f"Error connecting to Groundlight using API token '{self.api_token_prefix}...'"
                 f" at endpoint '{self.endpoint}'.  Endpoint might be invalid or unreachable? "
                 "Check https://status.groundlight.ai/ for service status."
+                f"Original Error was: {str(e)}"
```
brandon-groundlight, line 210: Oh, this was going to go into a different PR but I accidentally got it here. Tom and Tim noticed that we should immediately return the error at this point

src/groundlight/client.py
```diff
@@ -384,7 +388,7 @@ def create_detector(  # noqa: PLR0913
             # Create a detector in a specific group
             detector = gl.create_detector(
                 name="vehicle-counter",
```
sunildkumar, line 390: This should probably be named something different.

src/groundlight/client.py
```diff
@@ -384,7 +388,7 @@ def create_detector(  # noqa: PLR0913
...
-                query="How many vehicles are in the parking lot?",
+                query="Are there vehicles are in the parking lot?",
                 group_name="parking-monitoring",
                 patience_time=60.0
```
sunildkumar, line 393: Do we want to include patience time in the example? Probably not right?

src/groundlight/client.py
```diff
@@ -361,6 +362,9 @@ def create_detector(  # noqa: PLR0913
         """
         Create a new Detector with a given name and query.
 
+        Counting and Multiclass detectors are in Beta, and can be created through the
+        ExperimentalApi via the `create_counting_detector` and `create_multiclass_detector` methods.
```
sunildkumar, line 366: Is it obvious to a user how to find documentation on the experimental API? Should we include a link?

src/groundlight/client.py
brandon-groundlight: I don't see why not, we make it a legal parameter on the function

src/groundlight/client.py
brandon-groundlight: Does Docusaurus let you do that? Let me look

brandon-groundlight: Ooh, it does and it's really nice

src/groundlight/client.py
tyler-romero: Thanks! Good change

tyler-romero (approved):

sunildkumar (approved):

PR #317, opened 2025-02-21
Fix bug in wait_for_confident_result

CoreyEWood:
The `confidence_threshold` parameter defaults to `None`, and the docstring states that if it is `None` then the detector's confidence threshold will be used. However, it currently only fetches the detector's confidence threshold if the value for `image_query` is a string - if you pass it an `ImageQuery` object, `confidence_threshold` will still be `None` when passed to `iq_is_confident`, which will then error when trying to do `iq.result.confidence >= confidence_threshold`.

This fixes it so it will always fetch the detector's confidence threshold if the inputted `confidence_threshold` is None.


brandon-groundlight (approved):
Nice catch

PR #318, opened 2025-02-21
Model Unclear text label

brandon-groundlight:
Updating the public api spec to allow for modeling unclear text responses as null text


CoreyEWood (approved):
Looks good for the text result changes. Is a None value for `text` the official way to communicate "Unclear", or will that change in the future?

Also, is it fine for all the webhook changes to be in here too? Just wanted to double check.

brandon-groundlight:
> Looks good for the text result changes. Is a None value for `text` the official way to communicate "Unclear", or will that change in the future?

For the foreseeable future, None will be the official way to signal Unclear. We could conceivably add another field, but there's nothing else competing for the meaning of None

> Also, is it fine for all the webhook changes to be in here too? Just wanted to double check.
I'll be impressed if someone scoops our public-api.yaml before she merges her PR. It doesn't matter either way, but maybe for completeness I'll remove that from here

PR #319, opened 2025-02-22
Expose evaluations and metrics

brandon-groundlight:
Makes metrics and evaluations available


src/groundlight/experimental_api.py
```diff
@@ -997,3 +998,30 @@ def _download_and_save(url: str, output_path: str) -> bytes:
...
+
+    def get_detector_evaluation(self, detector: Union[str, Detector]) -> dict:
+        """
+        Get a specific evaluation for a detector
+
+        :param detector: the detector to get the evaluation for
+        :param evaluation_id: the evaluation id to get
+
+        :return: The DetectorEvaluation object
+        """
```
CoreyEWood, line 1010: I think the docstring here is wrong - it doesn't take an `evaluation_id`. It also returns a dict, should it be returning a `DetectorEvaluation` object?

src/groundlight/experimental_api.py
```diff
@@ -997,3 +998,30 @@ def _download_and_save(url: str, output_path: str) -> bytes:
...
+        obj = self.detectors_api.get_detector_evaluation(detector)
+        return obj.to_dict()
+
+    def get_detector_metrics(self, detector: Union[str, Detector]) -> dict:
+        """
+        Get the metrics for a detector
+
+        :param detector: the detector to get the metrics for
+
+        :return: The DetectorMetrics object
```
CoreyEWood, line 1022: Same here about the return type.

test/unit/test_metrics_and_evaluation.py
```diff
@@ -0,0 +1,33 @@
...
+        gl_experimental.add_label(iq, "NO")
+    metrics = gl_experimental.get_detector_metrics(det.id)
+    assert metrics["summary"] is not None
+    assert metrics["summary"]["num_ground_truth"] is not None
+    assert metrics["summary"]["num_current_source_human"] is not None
+    assert metrics["summary"]["class_counts"] is not None
+    assert metrics["summary"]["unconfident_counts"] is not None
+    assert metrics["summary"]["total_iqs"] is not None
+
+    # NOTE: this implicitly tests how quickly the evaluation is made available
```
CoreyEWood, line 26: Is there any worry that this test will be flaky due to evaluation sometimes taking too long?

CoreyEWood (approved):
Left some comments about docstrings/return types but functionality-wise this seems good.

src/groundlight/experimental_api.py
brandon-groundlight: Thanks for checking the docstrings, I left an older version in after I updated the function. Returning a dict is a little bit of a break from our usual format but this exposes the feature until we better formalize what should be in the evaluation.

test/unit/test_metrics_and_evaluation.py
brandon-groundlight: possibly, it's something I've got my eye one. If it gets in the way of the CICD pipeline I'll raise the amount of time we give to get an evaluation. But tests were passing on my local tests for the time being

PR #320, opened 2025-02-24
Add headers to webhook alerts

f-wright:
This PR allows the user to customize the headers field for their webhook alerts in addition to the payload. This will enable auth, which is required by a lot of different platforms.


spec/public-api.yaml
```diff
@@ -1242,6 +1242,8 @@ components:
       properties:
         template:
           type: string
+        headers:
```
brandon-groundlight, line 1245: Is there no openapi type for the headers? In the client I see that the headers are expected to be a `Optional[Dict[str, str]]`. Should we model that here?

generated/model.py
```diff
@@ -251,7 +253,7 @@ class MultiClassificationResult(BaseModel):
 class TextRecognitionResult(BaseModel):
     confidence: Optional[confloat(ge=0.0, le=1.0)] = None
     source: Optional[Source] = None
-    text: Optional[str] = Field(...)
+    text: str
```
brandon-groundlight, line 256: Can you double check how this got changed? I'm not seeing a corresponding change in the public-api.yaml file. My expectation is that this [line](https://github.com/groundlight/python-sdk/blob/0fa6244bd50ee0dcb49451c46014bdf159fd967e/spec/public-api.yaml#L1552) should mean that this is still optional. Is it possible that you generated the file with a work in progress yaml spec?

brandon-groundlight (commented):
Could you double check that the generated files are up to date? Also, it's worth considering modeling the data types for the headers. Everything else about the webhooks is modeled pretty well, and without looking at the zuuul side I believe all you would need to do is nest a header serializer inside the webhood serializer.

generated/model.py
f-wright: Not sure how it got changed, but regenerating seems to have solved it.

spec/public-api.yaml
f-wright: That's a good point -- I'll make the cloud service changes so that makes it into the spec.

generated/.openapi-generator/FILES
```diff
@@ -29,6 +29,11 @@ docs/ImageQueriesApi.md
 docs/ImageQuery.md
 docs/ImageQueryTypeEnum.md
 docs/InlineResponse200.md
+docs/InlineResponse2001.md
```
f-wright, line 32: This is unrelated to my change -- not sure if I should keep those changes out of the public-api.yaml file when generating?

brandon-groundlight: This is related to https://github.com/groundlight/python-sdk/pull/319. You can either ignore it here, or you can let me merge the other PR first and then you can merge main into your PR to get a cleaner diff

brandon-groundlight (approved):
LGTM

PR #325, opened 2025-03-05
Add new result types, make EDGE a source option, make `label` field of `LabelValue` nullable

CoreyEWood:
Updating generated code based on latest spec.


brandon-groundlight (approved):
go go generator

PR #330, opened 2025-03-13
generic requests using the api client

brandon-groundlight:
Provides a way to custom construct your own requests to groundlight. There's potential to clean it up when we deprecate the GroundlightApiClient class

### DIFF
--- a/src/groundlight/experimental_api.py
+++ b/src/groundlight/experimental_api.py
@@ -1064,24 +1064,9 @@ class ExperimentalApi(Groundlight):  # pylint: disable=too-many-public-methods
         headers["Accept"] = "application/json"
         return headers
 
-    def make_raw_rest_request(self, method: str, endpoint: str, body: Union[dict, None] = None) -> dict:
-        """
-        Make a raw REST request to the specified URL
-
-        :param method: the HTTP method to use
-        :param endpoint: the endpoint to send the request to - the url path appended after the
-            endpoint including a / at the beginging
-        :param body: the request body
-
-        :return: a dictionary containing the raw response
-        """
-        headers = self.get_raw_headers()
-        url = f"{self.api_client.configuration.host}{endpoint}"
-        response = requests.request(method, url, headers=headers, json=body)
-        return response.json()
-
     def make_generic_api_request(  # noqa: PLR0913 # pylint: disable=too-many-arguments
         self,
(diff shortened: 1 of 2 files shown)


src/groundlight/experimental_api.py
```diff
@@ -1046,3 +1048,68 @@ def get_detector_metrics(self, detector: Union[str, Detector]) -> dict:
...
+        headers = {}
+        # see generated/groundlight_openapi_client/api_client.py update_params_for_auth
+        headers["x-api-token"] = self.api_client.configuration.api_key["ApiToken"]
+        # We generate a unique request ID client-side for each request
+        headers["X-Request-Id"] = _generate_request_id()
+        headers["User-Agent"] = self.api_client.default_headers["User-Agent"]
+        headers["Accept"] = "application/json"
+        return headers
+
+    def make_raw_rest_request(self, method: str, endpoint: str, body: Union[dict, None] = None) -> dict:
```
robotrapta, line 1067: I'd leave this out unless we have a really good reason to include it.  One of these feels helpful, two starts to sound confusing.

src/groundlight/experimental_api.py
```diff
@@ -1046,3 +1048,68 @@ def get_detector_metrics(self, detector: Union[str, Detector]) -> dict:
...
+
+        :return: a dictionary containing the raw response
+        """
+        headers = self.get_raw_headers()
+        url = f"{self.api_client.configuration.host}{endpoint}"
+        response = requests.request(method, url, headers=headers, json=body)
+        return response.json()
+
+    def make_generic_api_request(  # noqa: PLR0913 # pylint: disable=too-many-arguments
+        self,
```
robotrapta, line 1084:
```suggestion
        self,
        *,
```

We could require the arguments to be named, not positionally-ordered.  Feels like a good practice for these methods with tons of arguments, although this one is mercifully more succinct than the openapi provided ones.

Is this the right syntax for that?

robotrapta (approved):
Thanks!  This is certainly helpful for what I'm trying to do.  The architecture here still feels wonky in that this stuff which feels like "standard way to format a request" is happening all over the place, and not in a single method like this which everything else is relying on.  I guess that's the price we pay for relying on the openapi generator, which has a bunch of other benefits.

PR #337, opened 2025-03-31
Bump version to 0.22.4 and generate client code from updated spec

CoreyEWood:
Bump version to 0.22.4.

This updates the spec to the point of #4798, to avoid removing the `Label` enum (which may cause issues, I'm not sure).

Spec changes include (as far as I can tell):
- additional fields related to Webhook models
- addition of bounding box mode components
- edge `report-metrics` endpoint
- `from_edge` field on results


pyproject.toml
```diff
@@ -9,7 +9,7 @@ packages = [
     {include = "**/*.py", from = "src"},
 ]
 readme = "README.md"
-version = "0.22.3"
+version = "0.22.4"
```
CoreyEWood, line 12: I think it's fine to bump the version here rather than in a separate PR? Let me know if that's not true though.

f-wright: It's fine to do here, we only bump it in a separate PR sometimes due to forgetting to put it in earlier :)

f-wright (approved):
LGTM!

PR #342, opened 2025-04-02
Make most evaluation results subfields nullable

paulina-positronix:
Make most evaluation results subfields nullable to avoid ApiTypeErrors when requesting evaluation results for count mode detectors. 
Most changes were automatically generated. The only manual changes are inside `spec/public_api.yaml`


f-wright (commented):
When you say "manual" changes, did you generate a spec and copy it over from zuuul, or change it directly? We keep those two versions of the API spec in sync, so we don't want to make actual manual changes here.

paulina-positronix:
> When you say "manual" changes, did you generate a spec and copy it over from zuuul, or change it directly? We keep those two versions of the API spec in sync, so we don't want to make actual manual changes here.

I did not know that -- will do it correctly now

PR #344, opened 2025-04-02
Update api spec

paulina-positronix:
- Update API spec from internal codebase
- Temporary fixes that use now missing model classes Label and Source


brandon-groundlight (approved):
I think this looks good to me

PR #346, opened 2025-04-03
Changing the detector per Leo's request

julia-groundlight:
Instead of "Is there an eagle", making the detector reflect a bit more of Groundlight's strengths


tyler-romero (approved):
LGTM!

PR #350, opened 2025-04-21
Adds summary table to submit_image_query method overview

brandon-groundlight:
Adds a table summary to the docs on the different usage of the submit image query methods


tyler-romero (approved):
Good idea

brandon-groundlight:
> Good idea

Admittedly it was Tom's idea

PR #353, opened 2025-04-28
Update API spec

paulina-positronix:
Update API spec to include optional count and multiclass mode evaluation results
Manually verified backwards compatible


brandon-groundlight (approved):
Looks good, I expect this change should not affect compatibility at all - older SDK versions should ignore additional fields and new versions can accept the lack of the new fields. Sounds like you confirmed this as well

PR #355, opened 2025-05-01
Bounding box mode detector creation

f-wright:
This PR adds a function to the experimental API to create bounding box mode detectors. This mode is not currently enabled for general usage, so the tests are disabled for now.

### DIFF
--- a/src/groundlight/experimental_api.py
+++ b/src/groundlight/experimental_api.py
@@ -903,10 +903,12 @@ class ExperimentalApi(Groundlight):  # pylint: disable=too-many-public-methods
             metadata=metadata,
         )
         detector_creation_input.mode = ModeEnum.COUNT
-        # TODO: pull the BE defined default
+
         if max_count is None:
-            max_count = 10
-        mode_config = CountModeConfiguration(max_count=max_count, class_name=class_name)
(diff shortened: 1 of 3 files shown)


src/groundlight/experimental_api.py
```diff
@@ -974,6 +975,79 @@ def create_multiclass_detector(  # noqa: PLR0913 # pylint: disable=too-many-argu
...
+            name=name,
+            query=query,
+            group_name=group_name,
+            confidence_threshold=confidence_threshold,
+            patience_time=patience_time,
+            pipeline_config=pipeline_config,
+            metadata=metadata,
+        )
+        detector_creation_input.mode = ModeEnum.BOUNDING_BOX
+        # TODO: pull the BE defined default
```
brandon-groundlight, line 1043: Do we need to pull it, or can you just submit with a null value here and serverside we interpret that as the default value

brandon-groundlight (approved):
Looks good, my main issue is with the TODO, but the code here right now is good

src/groundlight/experimental_api.py
f-wright: Good point, I modeled this off of count mode but I think we can just submit the null

PR #359, opened 2025-05-06
maintain Verb and Source class

brandon-groundlight:
Backwards compatibility for Verb and Source, fix issue with validation on result_type


generated/model.py
```diff
@@ -284,6 +304,21 @@ class VerbEnum(str, Enum):
     NO_QUERIES = "NO_QUERIES"
```
brandon-groundlight, line 306: The key part of this PR, restores two missing classes

sunildkumar (approved):

PR #361, opened 2025-05-12
Fix pip install typo for framegrab+youtube

tyler-romero:


brandon-groundlight (approved):
I want the old version to work so badly but pip doesn't give us everything we want

PR #363, opened 2025-05-15
Support creating text-mode detectors

tyler-romero:


brandon-groundlight (approved):
LGTM!
We have this test for the old text mode [here](https://github.com/groundlight/python-sdk/blob/d74fc46df3709a4bf5f74b0ee260e9214881cb8c/test/integration/test_groundlight.py#L308). I'm on the fence as to whether to remove it with this update, but regardless these changes look good

test/conftest.py
tyler-romero, line 1: conftests apply to all subdirs, so by relocating, we can reuse all these fixtures

PR #366, opened 2025-05-22
counting formal release

brandon-groundlight:
Move features that have had time to be battle tested in the experimental section into the main client without change other than minor doc updates


src/groundlight/client.py
```diff
@@ -362,9 +364,9 @@ def create_detector(  # noqa: PLR0913
         """
         Create a new Detector with a given name and query.
 
-        Counting and Multiclass detectors are in Beta, and can be created through the
-        ExperimentalApi via the :meth:`ExperimentalApi.create_counting_detector` and
-        :meth:`ExperimentalApi.create_multiclass_detector` methods.
+        Text and Bounding box detectors are in Beta, and can be created through the
```
f-wright, line 367: Should we indicate that these need to be enabled for your account, so people know to contact us if they'd like to create these detectors?

f-wright (approved):
LGTM! Do you want to bump the version and do a release for this?

honeytung (approved):
LGTM!

src/groundlight/client.py
brandon-groundlight: That should be included in the cloud response if they're not enabled already. Worth considering in the future if that message isn't cutting through

PR #374, opened 2025-06-11
Remove unnecessary call to delete all rules to reduce concurrent test flakiness

brandon-wada:
If two instances of tests were running on the same account in parallel, it's possible for delete all rules to collide with itself. We removed this previously for most of the other tests, one case of it appears to have been left in


timmarkhuff (approved):
LGTM

PR #376, opened 2025-06-16
Stop potential race condition in test

brandon-wada:
If Groundlight is still processing an image query and the detector is deleted, it can lead to undefined behavior


CoreyEWood (approved):
LGTM

PR #377, opened 2025-06-24
Update sample-applications.md

julia-groundlight:
Added images. I would have preferred to make it so that there could be up to 3 images in a row, but the table looked so ugly. So made it so it's a long list of images instead.


timmarkhuff:
LGTM

I made a few small comments.

I also fixed the <br> tag issue. Tests are rerunning now. They should pass this time.

timmarkhuff (approved):

jaredgroundlight (approved):

PR #381, opened 2025-08-20
Fix delete predictor test

roxanne-o:
Since we now throw a 410 GONE error on recently deleted soft-delete models, this test needs to be updated.


sunildkumar (approved):

brandon-wada (approved):
LGTM

PR #382, opened 2025-08-28
Add monthly usage

brandon-wada:
Allows users to see their month's usage via the SDK

### DIFF
--- a/test/integration/test_groundlight.py
+++ b/test/integration/test_groundlight.py
@@ -877,7 +877,7 @@ def test_delete_detector(gl: Groundlight):
     gl.delete_detector(detector)
 
     # Verify the detector is actually deleted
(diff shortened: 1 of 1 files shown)


src/groundlight/client.py
```diff
@@ -625,6 +628,28 @@ def list_image_queries(
             image_queries.results = [self._fixup_image_query(iq) for iq in image_queries.results]
         return image_queries
 
+    def get_month_to_date_usage(self) -> AccountMonthToDateInfo:
+        """
+        Get the account's month-to-date usage information including image queries (IQs),
```
sunildkumar, line 633: should this return other limits as well (e.g. throttle rate)?

sunildkumar (approved):

PR #383, opened 2025-08-29
adds ability to pin note in experimental api

brandon-wada:
Adds in the is_pinned parameter to the sdk function `create_note`


brandon-wada:
> lgtm - why experimental only?

We probably could move it over, it'd be nice to get other user's feedback before moving it though

PR #386, opened 2025-09-24
Enable configuring HTTP transport retry behavior and add `request_timeout` parameter to `get_detector`

CoreyEWood:
This PR includes two distinct changes.
- Adds a new parameter `http_transport_retries` to the `Groundlight` constructor, enabling the retry behavior to be modified when initializing a `Groundlight` instance. 
  - The retry behavior specified will apply to all API requests made through the `Groundlight` instance. 
  - Currently no way is provided to configure the retry behavior per-request, or to change it once the `Groundlight` instance has been initialized (this is technically possible but awkward). 
  - I added a test to verify that we can pass the parameter in and it gets passed through to the inner API client.
- Adds a `request_timeout` parameter to the `get_detector` method which enables changing the request timeout values for the request. 
  - An identical parameter was previously added for `submit_image_query`. Eventually I think we'll want to make this available for all methods, but for now we don't need that.
  - I added a test to verify that this parameter works properly.

These changes are intended to be utilized by the edge endpoint to improve behavior under poor network connectivity.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ python = ">=3.9,<4.0"
 python-dateutil = "^2.9.0"
 requests = "^2.28.2"
 typer = "^0.15.4"
-urllib3 = "^1.26.9"
+urllib3 = "^2.5.0"
 
 [tool.poetry.group.dev.dependencies]
 datamodel-code-generator = "^0.22.1"
--- a/src/groundlight/client.py
+++ b/src/groundlight/client.py
@@ -273,7 +273,7 @@ class Groundlight:  # pylint: disable=too-many-instance-attributes,too-many-publ
     def get_detector(
         self,
         id: Union[str, Detector],
-        request_timeout: Optional[float] = None,
+        request_timeout: Optional[Union[float, Tuple[float, float]]] = None,
     ) -> Detector:  # pylint: disable=redefined-builtin
         """
         Get a Detector by id.
@@ -675,7 +675,7 @@ class Groundlight:  # pylint: disable=too-many-instance-attributes,too-many-publ
         inspection_id: Optional[str] = None,
         metadata: Union[dict, str, None] = None,
         image_query_id: Optional[str] = None,
-        request_timeout: Optional[float] = None,
+        request_timeout: Optional[Union[float, Tuple[float, float]]] = None,
     ) -> ImageQuery:
         """
         Evaluates an image with Groundlight. This is the core method for getting predictions about images.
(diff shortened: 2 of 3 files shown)


src/groundlight/client.py
```diff
@@ -133,26 +134,31 @@ def __init__(
...
+            When not specified, checks the "DISABLE_TLS_VERIFY" environment variable (1=disable, 0=enable).
+            Certificate verification is enabled by default.
 
-                                       Warning: Only disable verification when connecting to a Groundlight Edge
-                                       Endpoint using self-signed certificates. For security, always keep
-                                       verification enabled when using the Groundlight cloud service.
+            Warning: Only disable verification when connecting to a Groundlight Edge Endpoint using self-signed
+            certificates. For security, always keep verification enabled when using the Groundlight cloud service.
+        :param http_transport_retries: Overrides urllib3 `PoolManager` retry policy for HTTP/HTTPS (forwarded to
+            `Configuration.retries`). Not the same as SDK 5xx retries handled by `RequestsRetryDecorator`.
```
CoreyEWood, line 152: Just corrected the formatting here.

src/groundlight/client.py
```diff
@@ -133,26 +134,31 @@ def __init__(
...
-                                       When not specified, checks the "DISABLE_TLS_VERIFY" environment variable
-                                       (1=disable, 0=enable). Certificate verification is enabled by default.
+            When not specified, checks the "DISABLE_TLS_VERIFY" environment variable (1=disable, 0=enable).
+            Certificate verification is enabled by default.
 
-                                       Warning: Only disable verification when connecting to a Groundlight Edge
-                                       Endpoint using self-signed certificates. For security, always keep
-                                       verification enabled when using the Groundlight cloud service.
+            Warning: Only disable verification when connecting to a Groundlight Edge Endpoint using self-signed
+            certificates. For security, always keep verification enabled when using the Groundlight cloud service.
```
CoreyEWood, line 150: Just corrected the formatting on these.

pyproject.toml
```diff
@@ -21,7 +21,7 @@ python = ">=3.9,<4.0"
 python-dateutil = "^2.9.0"
 requests = "^2.28.2"
 typer = "^0.15.4"
-urllib3 = "^1.26.9"
+urllib3 = "^2.5.0"
```
CoreyEWood, line 24: This upgrade is necessary for the `retries` parameter on the `Configuration` to actually be respected when requests are made. 2.5.0 is the most recent release - likely the fix was made earlier, but I don't see a reason not to update to the most recent.

PR #420, opened 2026-03-27
feat: add PrimingGroup and MLPipeline SDK methods

brandon-wada:
## Summary

Adds SDK methods for the new PrimingGroup public API. Priming Groups are groupings of detectors that share a common semantic query and can utilize a common base model. In the future, priming groups may allow for cross detector learning and other multi-task and meta-learning techniques. 

In introducing priming groups, we also expose MLPipelines. Every detector has one active MLPipeline but constantly runs multiple MLPipelines under the hood that are compared against the top performing active model. Whenever Groundlight determines that there is a better performing model, we will switch that model to be the active model.

Currently, a priming group can be created from an existing and trained MLPipeline, which one can get given an active detector. The new PrimingGroup will use a frozen version of the MLPipeline for the first inference on a new detector created using the PrimingGroups. If shadow pipelines are disallowed on the PrimingGroup, that frozen model will be used indefinitely by the detector. Otherwise, the frozen model will be used until one of the other MLPipelines on the detector is evaluated to be a better fit.

We add the following methods in this PR
- `list_detector_pipelines(detector)` — list all MLPipelines for a detector
- `create_priming_group(name, source_ml_pipeline_id, ...)` — create a PrimingGroup seeded from an existing trained model
- `list_priming_groups()` — list all priming groups owned by this account
- `get_priming_group(priming_group_id)` — retrieve by ID
- `delete_priming_group(priming_group_id)` — soft-delete

We also add new pydantic models: `MLPipeline`, `PrimingGroup` in `generated/model.py`.

## Usage

\`\`\`python
gl = Groundlight()

pipelines = gl.list_detector_pipelines(detector)
active = next(p for p in pipelines if p.is_active_pipeline)

pg = gl.create_priming_group(
    name="door-detector-primer",
    source_ml_pipeline_id=active.id,
    canonical_query="Is the door open?",
    disable_shadow_pipelines=True,
)

new_detector = gl.create_detector(
    name="new-door-detector",
    query="Is the door open?",
    priming_group_id=pg.id,
)
\`\`\`


test/unit/test_priming_groups.py
```diff
@@ -0,0 +1,170 @@
...
+    assert fetched.id == pg.id
+    assert fetched.name == pg.name
+
+    gl_experimental.delete_priming_group(pg.id)
+
+
+@pytest.mark.expensive
+def test_get_priming_group_unknown_raises(gl_experimental: ExperimentalApi):
+    with pytest.raises(NotFoundError):
+        gl_experimental.get_priming_group("pgp_doesnotexist000000000000")
```
timmarkhuff: I think this is a typo? pg instead of pgp

timmarkhuff:
From Claude:
1. Both list_detector_pipelines and list_priming_groups only return the first page. If there are more items than the default page size, results are silently truncated. Consider at least documenting this, or exposing page/page_size parameters like list_rules does.
2. Every test in test/unit/test_priming_groups.py hits real API endpoints via the gl_experimental and detector fixtures (which create real detectors against the server). This file should be at test/integration/test_priming_groups.py.
3. parse_obj is deprecated in Pydantic v2. Other code in the same file already uses model_validate (e.g., line 315). All of these should use model_validate consistently.

timmarkhuff:
The PR description says "gl = Groundlight()" but it should say "gl = ExperimentalApi"

brandon-wada:
A couple of these will require backend changes, but since we have multiple consumers of this now I think we should merge this and file those as bugs to fix soon. Specifically
> Both list_detector_pipelines and list_priming_groups only return the first page. If there are more items than the default page size, results are silently truncated. Consider at least documenting this, or exposing page/page_size parameters like list_rules does.


> It seems there is no way to connect the dots between a detector and it's related priming group. Below is a detector that I created on a particular priming group, but there is no way (that I can find) to know that they are connected.
Detectors know their Priming group, we need to do a quick BE fix to expose that

> The primed detector doesn't get the mature weights downloaded to run on edge
Pushing priming to the edge is a purely backend issue

PR #428, opened 2026-04-27
ExperimentalAPI CLI

timmarkhuff:
1. adds support for `ExperimentalApi` methods in CLI
2. converts return values into human-readable strings
3. introduces a mechanism to hide certain functions from the CLI (either because they cannot return a value that is human readable, or because they just don't belong in the CLI)
4. entirely removes certain functions from the `ExperimentalApi` that needed to be deprecated. 
5. organizes CLI functions into groupings and alphabetizes them


src/groundlight/cli.py
```diff
@@ -38,14 +120,22 @@ class FakeClass:
...
+        if is_experimental:
+            print(
+                f"Warning: '{fake_method.__name__}' is an experimental command and may change without notice.",
+                file=sys.stderr,
+            )
+        gl = ExperimentalApi()
+        bound_method = getattr(gl, fake_method.__name__)
+        result = bound_method(*args, **kwargs)
+        if result is not None:
+            print(_format_result(result))
```
brandon-wada, line 132: Heavily biased as the one who wrote the original, but there should be more comments around this one line. This is the magic line that makes the whole thing happen

src/groundlight/cli.py
```diff
@@ -71,13 +166,116 @@ def wrapper(*args, **kwargs):
...
+    "Image Queries",
+    "ML Pipelines & Priming",
+    "Notes",
+    "Utilities",
+    "Other",
+]
+
+# Maps method names to their rich_help_panel group label for the CLI help output.
+# Applies to both stable and experimental commands.
+_COMMAND_GROUPS: dict[str, str] = {
```
brandon-wada, line 188: I think this adds maintenance overhead, new methods need to belong to a group. The original design tried to avoid that additional overhead, but in the age of AI maybe we don't care

src/groundlight/cli.py
timmarkhuff: Done. I added a comment.

src/groundlight/cli.py
timmarkhuff: I added a test to ensure that each method has a CLI group. It will be an extra step, but a worthwhile extra step that cannot be forgotten.

brandon-wada (approved):
Looks good!

PR #430, opened 2026-05-05
feat: plumb detector_mode through create_priming_group

brandon-wada:
Updates priming group tests according to new validation constraints.


generated/docs/DetectorModeEnum.md
```diff
@@ -0,0 +1,12 @@
+# DetectorModeEnum
+
+* `BINARY` - BINARY * `COUNT` - COUNT * `MULTI_CLASS` - MULTI_CLASS * `TEXT` - TEXT * `BOUNDING_BOX` - BOUNDING_BOX
```
f-wright, line 3: I know this doesn't go in this PR, but it seems like it'd make more sense for these to be in alphabetical order. Right now they seem to be in order of when they were added, but that's kind of odd since we don't really use count and text mode all that much anymore

generated/docs/PrimingGroupCreationInputRequest.md
```diff
@@ -7,6 +7,7 @@ Name | Type | Description | Notes
 ------------ | ------------- | ------------- | -------------
 **name** | **str** | Name for the new priming group. | 
 **source_ml_pipeline_id** | **str** | ID of an MLPipeline owned by this account whose trained model will seed the priming group. | 
+**detector_mode** | **bool, date, datetime, dict, float, int, list, str, none_type** | Detector mode this priming group is intended for (BINARY, MULTI_CLASS, etc.). Must match the mode supported by the source MLPipeline&#39;s pipeline_config.  * &#x60;BINARY&#x60; - BINARY * &#x60;COUNT&#x60; - COUNT * &#x60;MULTI_CLASS&#x60; - MULTI_CLASS * &#x60;TEXT&#x60; - TEXT * &#x60;BOUNDING_BOX&#x60; - BOUNDING_BOX |
```
f-wright, line 10: Why does this have so many type options? Do we need to tighten this on the back end?

generated/groundlight_openapi_client/model/detector_mode_enum.py
```diff
@@ -0,0 +1,286 @@
+"""
+    Groundlight API
+
+    Groundlight makes it simple to understand images. You can easily create computer vision detectors just by describing what you want to know using natural language.  # noqa: E501
+
+    The version of the OpenAPI document: 0.18.2
+    Contact: support@groundlight.ai
```
f-wright, line 7: This is out of date, but maybe we don't care

generated/groundlight_openapi_client/model/null_enum.py
```diff
@@ -0,0 +1,282 @@
...
+    datetime,
+    file_type,
+    none_type,
+    validate_get_composed_info,
+    OpenApiModel,
+)
+from groundlight_openapi_client.exceptions import ApiAttributeError
+
+
+class NullEnum(ModelSimple):
```
f-wright, line 32: Maybe this is expected in the generated code, but seems weird to me? Why do we need this `NullEnum` class?

f-wright: Especially because it looks like we have a separate `BlankEnum` already?

generated/test/test_detector_mode_enum.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def setUp(self):
+        pass
+
+    def tearDown(self):
+        pass
+
+    def testDetectorModeEnum(self):
+        """Test DetectorModeEnum"""
+        # FIXME: construct object with mandatory attributes with example values
```
f-wright, line 29: Does a FIXME comment indicate we should do something about this before merging?

test/integration/test_priming_groups.py
```diff
@@ -88,6 +88,7 @@ def test_create_priming_group(gl_experimental: ExperimentalApi, detector):
     pg = gl_experimental.create_priming_group(
         name=f"test-primer-{detector.id}",
         source_ml_pipeline_id=trained.id,
+        detector_mode=ModeEnum.BINARY,
```
f-wright, line 120: Should we have any tests for the multiclass specific num classes argument usage?

f-wright (commented):
I'm a little confused about some of what's happening here, but if you think all of this is expected I'm happy to trust your judgement on this

test/integration/test_priming_groups.py
brandon-wada: Added `test_create_priming_group_multiclass_populates_num_classes` in 7c8400c5 — creates a 3-class detector, trains it, seeds a priming group with `detector_mode=ModeEnum.MULTI_CLASS`, and asserts `pg.num_classes == 3`. Marked `@pytest.mark.expensive` like the rest of the create-flow tests. Couldn't run it locally (the integ token I had has been revoked) — will rely on CI.

f-wright (approved):

PR #442, opened 2026-06-18
Add ask_vlm_verify in ExperimentalApi for cloud VLM alert verification

srnangi:
## What

Adds `ExperimentalApi.ask_vlm_verify(media, query, model_id)` for VLM-based alert verification. Calls `POST /v1/vlm-verifications` on the Groundlight cloud (AWS Bedrock) and returns a **generated** `VlmVerification` model with a nested `result` (`verdict` = `YES`/`NO`/`UNSURE`, `confidence`, `reasoning`) and `cost` (`input_tokens`, `output_tokens`, `total_cost_usd`).

Pairs with the janzu PR (zuuul#6519). No local inference — the VLM runs entirely in the cloud.

## Generated from the OpenAPI spec

Per review feedback, the response model is **not** hand-written — it follows the SDK codegen process:

- The endpoint + schemas (`VlmVerification`, `VlmVerificationResult`, `VlmVerificationCost`, `VerdictEnum`) were added to `spec/public-api.yaml`, synced from zuuul#6519's regenerated `public-api.yaml`.
- `make generate` regenerates both `generated/model.py` (the Pydantic `VlmVerification` the SDK returns) and the `groundlight_openapi_client` package (`vlm_verifications_api.py` + model classes).

## Experimental surface

The method lives on `ExperimentalApi`, not the core `Groundlight` client. It returns the generated `VlmVerification` model directly.

## API

- `media` accepts **1–8 images** (single or list). Accepts numpy BGR arrays, PIL Images, bytes, BytesIO/BufferedReader, or filename strings — encoded via the existing `parse_supported_image_types` utility.
- POSTs `multipart/form-data`: `media` parts as `image/jpeg` files; `query` and `model_id` as form fields (not URL params, so the prompt never leaks into access logs).
- `model_id` is a **friendly alias** (e.g. `"gpt-5.4"`, `"claude-sonnet-4.5"`). The server maps it to the real Bedrock model id and returns 400 on an unknown alias. Defaults to the server-configured default.
- The request goes through the **generated client** (`VlmVerificationsApi.submit_vlm_verification`) — the same pattern as `submit_image_query` — and the response is validated into the generated `VlmVerification` model via `model_validate`. The generated multipart helper derives each `media` part's filename/content-type from the file's `.name`, so `ByteStreamWrapper` (the SDK's existing image-bytes carrier) gains an optional `name`.

## Usage

```python
from groundlight import ExperimentalApi

gl = ExperimentalApi()

# Single image
result = gl.ask_vlm_verify(frame, query="Is there a fire in this image?")
if result.result.verdict == "YES":
    emit_alert()

# Full frame + cropped ROI — describe each in the query
result = gl.ask_vlm_verify(
    media=[full_frame, roi_crop],
    query="Image 1 is the full camera frame; image 2 is the cropped region "
          "a detector flagged. Is there really a fire?",
)
print(result.result.confidence, result.result.reasoning, result.cost.total_cost_usd)
```

## Changes

- `spec/public-api.yaml`: add `/v1/vlm-verifications` path + `VlmVerification` / `VlmVerificationResult` / `VlmVerificationCost` / `VerdictEnum` schemas
- `generated/`: regenerated client (`generated/model.py` + `groundlight_openapi_client` VLM api/model files) via `make generate`
- `src/groundlight/experimental_api.py`: `ask_vlm_verify` method (calls the generated `VlmVerificationsApi`)
- `src/groundlight/images.py`: `ByteStreamWrapper` gains an optional `name` (needed for multipart file parts)
- `src/groundlight/cli.py`: register `ask_vlm_verify` as an experimental CLI command under a new "VLM Verification" group (`groundlight exp ask-vlm-verify MEDIA QUERY`). **Note: only a single image is supported via the CLI for now** — the `media` Union collapses to a `str` filepath. The Python API (`ExperimentalApi.ask_vlm_verify`) keeps full 1–8 image support.
- `test/unit/test_ask_vlm_verify.py` + `test/unit/test_cli.py`: unit tests (mocked HTTP) and a CLI registration test

This PR makes **no changes** to `client.py` or `__init__.py`.

## Testing

The endpoint isn't deployed yet, so tests are mocked (no live server). Coverage:

- Server JSON correctly parsed into a `VlmVerification` (nested `result.verdict`/`result.confidence`, `cost.total_cost_usd`)
- numpy image encoded to JPEG and sent as a multipart `media` part
- `query`/`model_id` sent as form fields, not URL params (prompt must not appear in logs)
- empty list and more than 8 media items each raise `ValueError` before any network call
- URL path correct (`/device-api/v1/vlm-verifications`)

## Follow-ups before merge / deploy

- **Depends on zuuul#6519** — the SDK spec was synced from that (unmerged) branch. Don't merge this before zuuul#6519, and re-sync `spec/public-api.yaml` if the server's final schema changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/groundlight/client.py
```diff
@@ -1089,6 +1107,118 @@ def ask_async(  # noqa: PLR0913 # pylint: disable=too-many-arguments
...
+        :param media: One image or a list of up to 8 images.  Accepted formats per image:
+
+            - filename (string) of a JPEG/PNG file
+            - raw bytes or BytesIO / BufferedReader
+            - numpy array (H, W, 3) in BGR order (OpenCV convention)
+            - PIL Image
+
+        :param query: Natural-language prompt describing the media and what to verify,
+            e.g. ``"Is there a fire visible in the image? Reason step by step."``
+        :param model_id: Friendly alias of the VLM to use, e.g.
```
srnangi: Add model names from the model registry available to use here.

test/unit/test_ask_vlm.py
```diff
@@ -0,0 +1,127 @@
...
+        "model_id": model_id,
+        "result": {"verdict": verdict, "confidence": confidence, "reasoning": reasoning},
+        "cost": {"input_tokens": 400, "output_tokens": 80, "total_cost_usd": 0.0015},
+    }
+    resp.raise_for_status = MagicMock()
+    return resp
+
+
+class TestAskVlm:
+    @patch("groundlight.client.requests")
```
srnangi: Should we add test for timeout?

src/groundlight/client.py
```diff
@@ -1089,6 +1110,130 @@ def ask_async(  # noqa: PLR0913 # pylint: disable=too-many-arguments
...
+        self,
+        media: Union[
+            "np.ndarray",
+            List["np.ndarray"],
+            str,
+            bytes,
+            "Image.Image",
+            BytesIO,
+            BufferedReader,
+        ],
```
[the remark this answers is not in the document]
srnangi: Fixed — `List` is now typed as `List[Union["np.ndarray", str, bytes, "Image.Image", BytesIO, BufferedReader]]` so it accurately reflects all supported per-item types. _🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/groundlight/client.py
```diff
@@ -1089,6 +1110,130 @@ def ask_async(  # noqa: PLR0913 # pylint: disable=too-many-arguments
...
+            ``"UNSURE"``), ``confidence``, ``reasoning``, and token cost fields.
+        :raises ValueError: If more than ``MAX_VLM_MEDIA_ITEMS`` (8) images are supplied.
+        :raises requests.HTTPError: On non-2xx response (400 for invalid model alias
+            or undecodable image bytes; 502 if the upstream VLM is unavailable).
+        """
+        # Normalise: single image → list
+        if not isinstance(media, list):
+            media = [media]
+        if len(media) > MAX_VLM_MEDIA_ITEMS:
+            raise ValueError(f"ask_vlm supports at most {MAX_VLM_MEDIA_ITEMS} media items.")
```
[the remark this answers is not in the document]
srnangi: Added `if not media: raise ValueError("ask_vlm requires at least one media item.")` before the encoding loop. _🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/groundlight/client.py
```diff
@@ -1089,6 +1110,130 @@ def ask_async(  # noqa: PLR0913 # pylint: disable=too-many-arguments
...
+        if model_id:
+            form_data["model_id"] = model_id
+
+        headers = {
+            "x-api-token": self.api_client.configuration.api_key["ApiToken"],
+            "X-Request-Id": f"ask_vlm_{int(time.time() * 1000)}",
+            "x-sdk-language": "python",
+        }
+
+        url = f"{self.endpoint}/v1/vlm-verifications"
```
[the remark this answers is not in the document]
srnangi: Fixed with `base = re.sub(r"/v\d+$", "", self.endpoint)` before appending `/v1/vlm-verifications`. Handles both `.../device-api` → `.../device-api/v1/...` and `.../v1` → `.../v1/...` correctly. _🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/groundlight/client.py
```diff
@@ -1089,6 +1110,130 @@ def ask_async(  # noqa: PLR0913 # pylint: disable=too-many-arguments
...
+
+        # query and model_id are sent as multipart form fields (not query-string
+        # params): the prompt can be long and must not end up in URLs or access logs.
+        form_data: dict[str, str] = {"query": query}
+        if model_id:
+            form_data["model_id"] = model_id
+
+        headers = {
+            "x-api-token": self.api_client.configuration.api_key["ApiToken"],
+            "X-Request-Id": f"ask_vlm_{int(time.time() * 1000)}",
```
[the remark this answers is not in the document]
srnangi: Switched to `time.time_ns()` — nanosecond precision makes collisions negligible without adding dependencies. _🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/groundlight/client.py
```diff
@@ -1089,6 +1110,130 @@ def ask_async(  # noqa: PLR0913 # pylint: disable=too-many-arguments
...
+                      "a detector flagged. Is there really a fire?",
+            )
+            print(result.confidence, result.reasoning)
+
+        :param media: One image or a list of up to 8 images.  Accepted formats per image:
+
+            - filename (string) of a JPEG/PNG file
+            - raw bytes or BytesIO / BufferedReader
+            - numpy array (H, W, 3) in BGR order (OpenCV convention)
+            - PIL Image
```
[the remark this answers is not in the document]
srnangi: Clarified in the docstring: bytes/streams accept any common image format (JPEG, PNG, WEBP) and the server normalises to JPEG server-side via `ensure_jpeg_format`. _🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/groundlight/client.py
```diff
@@ -1089,6 +1111,136 @@ def ask_async(  # noqa: PLR0913 # pylint: disable=too-many-arguments
...
+            raise ValueError("ask_vlm requires at least one media item.")
+        if len(media) > MAX_VLM_MEDIA_ITEMS:
+            raise ValueError(f"ask_vlm supports at most {MAX_VLM_MEDIA_ITEMS} media items.")
+
+        # Convert each image to JPEG bytes via the existing SDK utility
+        media_files: list[tuple[str, tuple[str, bytes, str]]] = []
+        for i, img in enumerate(media):
+            stream = parse_supported_image_types(img)
+            jpeg_bytes = stream.read()
+            media_files.append(("media", (f"image_{i}.jpg", jpeg_bytes, "image/jpeg")))
```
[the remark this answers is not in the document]
srnangi: Fixed: updated the  docstring to accurately state that filename inputs accept only JPEG/PNG (not WEBP), and that raw `bytes`/`BytesIO`/`BufferedReader` are forwarded as-is — the server's `ensure_jpeg_format` validates by decoding regardless of declared content-type, so PNG/WEBP bytes all work correctly. Also corrected the inline comment from "Convert each image to JPEG bytes" to reflect that only numpy/PIL paths actually convert.

test/unit/test_ask_vlm.py
```diff
@@ -0,0 +1,94 @@
...
+
+def test_url_has_correct_path(gl: Groundlight):
+    """sanitize_endpoint_url strips the trailing slash from self.endpoint, so the path
+    must include a leading '/' — without it the URL becomes '...device-apiv1/...'."""
+    with mock.patch("groundlight.client.requests") as mock_requests:
+        mock_requests.post.return_value = _mock_response()
+        gl.ask_vlm(media=_FAKE_JPEG, query="test")
+
+    args, _ = mock_requests.post.call_args
+    assert "/device-api/v1/vlm-verifications" in args[0]
```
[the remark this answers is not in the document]
srnangi: Fixed: added test_url_no_version_duplication_for_versioned_endpoint which creates a Groundlight instance with endpoint='http://test-server/v1' and asserts the resulting URL does not contain /v1/v1/ and ends with /v1/vlm-verifications.

src/groundlight/client.py
```diff
@@ -73,6 +76,25 @@ class EdgeNotAvailableError(GroundlightClientError):
     """Raised when an edge-only method is called against a non-edge endpoint."""
 
 
+MAX_VLM_MEDIA_ITEMS = 8
+
+
+@dataclass
```
timmarkhuff: This should be a Pydantic model. That matches the pattern we have established in this SDK.

Consistency aside, I think there will also be a problem with the CLI if we use a dataclass. The CLI has special logic to pretty print objects, and I don't believe it has any handling for dataclasses.

src/groundlight/client.py
```diff
@@ -73,6 +76,25 @@ class EdgeNotAvailableError(GroundlightClientError):
...
+class VLMVerificationResult:  # pylint: disable=too-many-instance-attributes
+    """Result of a VLM-based alert verification via the Groundlight cloud API."""
+
+    id: str
+    query: str
+    model_id: str
+    verdict: str  # "YES" | "NO" | "UNSURE"
```
timmarkhuff: Should we call this label or result instead of verdict? That would match established patterns better. But maybe this is meant to be intentionally different? 

It feels like we should be making use of existing objects, like `Result`. Right now, this class has verdict and confidence as two separate, seemingly unrelated concepts, but shouldn't they be grouped together as a `Result`?

src/groundlight/client.py
```diff
@@ -73,6 +76,25 @@ class EdgeNotAvailableError(GroundlightClientError):
...
+    confidence: float  # 0.0–1.0
+    reasoning: str
+    created_at: str
+    input_tokens: Optional[int] = None
+    output_tokens: Optional[int] = None
+    total_cost_usd: Optional[float] = None
```
timmarkhuff: I've heard it argued that `float` is not great for representing currency due to rounding issues. `int` or `Decimal` might be better.

src/groundlight/client.py
```diff
@@ -1089,6 +1111,139 @@ def ask_async(  # noqa: PLR0913 # pylint: disable=too-many-arguments
             inspection_id=inspection_id,
         )
 
+    def ask_vlm(  # pylint: disable=too-many-locals
+        self,
+        media: Union[
+            "np.ndarray",
```
timmarkhuff: I don't think we should be double quoting the type annotations here. Elsewhere in this same file we do not.

src/groundlight/client.py
timmarkhuff: We shouldn't be hand writing a class like this, should we? Most classes like this are autogenerated by the API spec.

timmarkhuff (changes requested):
I think there are some consistency issues with this PR that I have noted in my comments. We'll want Brandon to weigh in on these too.

src/groundlight/client.py
srnangi: Fixed: converted VLMVerificationResult from a dataclass to a pydantic BaseModel, matching the established SDK pattern and enabling proper CLI pretty-printing via model_dump_json.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/groundlight/client.py
srnangi: Fixed: removed the string forward-reference quotes from np.ndarray and Image.Image in the ask_vlm signature — consistent with every other method in client.py that uses the same optional_imports (lines 717, 881, etc.).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/groundlight/client.py
srnangi: Not changed — VLM verification outputs YES/NO/UNSURE which is intentionally different from image-query binary/multiclass labels. Grouping verdict+confidence into a nested Result felt like premature reuse: the existing Result objects are generated from the detector image-query spec and carry detector-specific semantics. Keeping them flat and top-level on VLMVerificationResult avoids that coupling. Happy to revisit if there is a strong preference for nesting, or once the spec is auto-generated.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/groundlight/client.py
srnangi: Left as float — total_cost_usd is an informational/display field echoed back from the server; the SDK performs no arithmetic on it, so Decimal's precision guarantees are not needed here. If the server ever returns this value in a way that requires client-side summing or billing math, switching to Decimal would be the right call.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/groundlight/client.py
srnangi: Acknowledged — the ideal end-state is for VLMVerificationResult to be generated from the OpenAPI spec (alongside VlmVerification in zuuul). The janzu endpoint is not deployed yet, so wiring up spec generation is a follow-on task. For now the hand-written Pydantic BaseModel gives correct CLI behaviour and the same interface, and can be replaced by the generated model once the spec is stable.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/groundlight/cli.py
```diff
@@ -232,6 +233,8 @@ def wrapper(*args, **kwargs):
     "create_priming_group": "ML Pipelines & Priming",
     "get_priming_group": "ML Pipelines & Priming",
     "delete_priming_group": "ML Pipelines & Priming",
+    # VLM Verification
+    "ask_vlm": "VLM Verification",
```
paulina-positronix: On the one hand, similar to my comments on the `zuuul` PR, if this is specific to verification then maybe the method should be `ask_vlm_verify` or similar, since the returned result will be verification-specific. On the other hand, I kinda like `ask_vlm` because it could be also used to just get the answer in the first place. But maybe we want to discourage such use due to potential for high cost. Hm... I'm of two minds about it, what do you think?

srnangi:
> I think there are some consistency issues with this PR that I have noted in my comments. We'll want Brandon to weigh in on these too.

Thanks @timmarkhuff for the comments. My claude agent was on auto-reply mode for this PR, so it replied on my behalf without me looking at it, sorry about that. I'll take a look at your comments again carefully.

src/groundlight/client.py
```diff
@@ -73,6 +76,24 @@ class EdgeNotAvailableError(GroundlightClientError):
...
+class VLMVerificationResult(BaseModel):
+    """Result of a VLM-based alert verification via the Groundlight cloud API."""
+
+    id: str
+    query: str
+    model_id: str
+    verdict: str  # "YES" | "NO" | "UNSURE"
+    confidence: float  # 0.0–1.0
+    reasoning: str
+    created_at: str
```
timmarkhuff: I think this doesn't match how we handle created_at elsewhere in this repo. I think we use datetime or something more specific than str.

timmarkhuff:
A few more things that just occurred to me:
1. this method was put directly under `Groundlight`. Typically we put new methods like this under `ExperimentalApi` first, and then in theory they get promoted later once they are battle tested
2. I'm not 100% sure if this method will work out of the box with the CLI. It might work if you provide a single image, not sure about multiple. This isn't a huge issue, and I would be happy to write a follow on PR to make sure it works in the CLI.

src/groundlight/client.py
brandon-wada, line 1: Should probably go in experimental

src/groundlight/client.py
```diff
@@ -73,6 +76,24 @@ class EdgeNotAvailableError(GroundlightClientError):
     """Raised when an edge-only method is called against a non-edge endpoint."""
 
 
+MAX_VLM_MEDIA_ITEMS = 8
+
+
+class VLMVerificationResult(BaseModel):
```
brandon-wada: Is there a reason we aren't modeling this in Swagger?

src/groundlight/cli.py
brandon-wada, line 174: Do you want this in the cli? Definitely needs tests if so

brandon-wada (changes requested):
This is a complete break from the OpenApi modeling that underpins most of the SDK. We can leave this on branch and test with it, but once we're confident that this is the feature we want then we should model the endpoint

Alternately, we have the `make_generic_api_request` method, we could just leverage that for the moment

paulina-positronix (commented):
Get Brandon's help to learn how to update the SDK using the right swagger tools etc. and otherwise I have one comment/musing on the naming of the method.

src/groundlight/cli.py
timmarkhuff: This got me thinking about the name `VLMVerificationResult`. Maybe something like `VlmResult` would be better. I get that we might encourage people to only use this for verification because it's expensive, but it isn't inherently a "verification"; it's only a verification if you use it that way. 

Also, this is an opinionated nit, but I think for class names we should use `Vlm..` rather than `VLM...`. The argument is that Pascal case is about word boundaries, not capitalization. We already follow this pattern in classes like `ExperimentalApi` so maybe it is worth continuing that pattern.

srnangi: These are both great points, and they made me think about the API design a bit more. I'm leaning towards making this more specific and renaming it to `ask_vlm_verify` and use `VlmVerificationResult`.

The main reason is that, in Zuuul, we built the Bedrock client and the `vlm-verification` endpoint primarily for verification tasks, with a system prompt and answer parsing (verdict, confidence, reasoning) specifically tailored for alert verification. While we could make it more generic by allowing callers to pass their own system prompts and get raw answers, I think it's better to keep this endpoint focused on verification for now to get structured output. This also helps discourage using it for general-purpose VLM calls, given the relatively high cost.

Additionally, I've added restrictions in the Zuuul endpoint so that only certain customer types can access it. Given that context, I'd prefer to optimize for the specific use case we're building today and generalize it later if and when we have a clear need.

src/groundlight/cli.py
srnangi: I see this to be primarily used by vision-engine service or to run some experiments, so I dont think we have a CLI use case here. So, I'm not adding it in this PR.

src/groundlight/experimental_api.py
```diff
@@ -164,6 +168,115 @@ def create_note(
...
+        form_data = {"query": query}
+        if model_id:
+            form_data["model_id"] = model_id
+
+        headers = {
+            "x-api-token": self.configuration.api_key["ApiToken"],
+            "X-Request-Id": _generate_request_id(),
+        }
+
+        url = f"{self.endpoint}/v1/vlm-verifications"
```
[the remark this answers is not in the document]
srnangi: This is how it is done with the other methods too. We have an endpoint with /v1/vlm-verifications

src/groundlight/cli.py
srnangi: Added CLI and tests, but it only supports 1 image for now.

srnangi:
Updated the PR with OpenAPI generated SDK and moved `ask_vlm_verify` to experimental API.

spec/public-api.yaml
```diff
@@ -962,6 +962,69 @@ paths:
...
+        `query` field and an optional `model_id` field.
+
+        The `query` describes what each image is and what to look for — the server makes
+        no assumptions about the images' meaning. Images are presented to the model
+        labeled `Image 1`, `Image 2`, ... in upload order, so the query can reference
+        them (e.g. "Image 1 is the full frame; image 2 is the cropped ROI ...").
+
+        (Video parts are planned but not yet supported and are rejected.)
+
+        Requires `ENABLE_BEDROCK_VLM_ACCESS` (enabled for Standard_Internal and SciDuck accounts) and accepted terms of service.
```
brandon-wada: We're leaking documentation to the public sdk here

spec/public-api.yaml
```diff
@@ -2338,6 +2401,79 @@ components:
...
+    VlmVerificationCost:
+      type: object
+      properties:
+        input_tokens:
+          type: integer
+          nullable: true
+        output_tokens:
+          type: integer
+          nullable: true
+        total_cost_usd:
```
brandon-wada, line 2441: Should this be here?

brandon-wada (approved):
We should clean up the docstrings in the api spec, but otherwise looks good

spec/public-api.yaml
srnangi: I didn't get your comment here, I'm also reporting the costs along with the verdict, so we need this. 
But, are you questioning something else?

timmarkhuff (approved):

PR #445, opened 2026-07-09
feat: automatically rotate SDK API tokens

timmarkhuff:
## Overview

This PR adds SDK support for automatic token rotation. The server-side support for token rotation already exists, however the frontend changes that allow a rotating token to be created aren't merged yet, so there will be no immediate effect of releasing this version; `Groundlight` clients will continue to function as they always have until Reef allows users to start creating rotating tokens.

## Summary

- Automatically rotate SDK API tokens when the configured token's identity has a non-null `token_ttl`.
- Non-rotating configured tokens (`token_ttl=null`) are used as-is: no mint, no on-disk cache, no background refresh thread (same behavior as pre-rotation `Groundlight`).
- For rotating tokens: mint a working child (server-authoritative lifetime; create omits client `expires_at`), and refresh every `token_ttl / 30`.
- Working token state is stored in a per-configured-token-snippet JSON cache on disk (current + previous), updated under a file lock so multiple processes on the same machine share one rotation chain instead of each minting independently.
- On refresh, revoke the old `previous` (best effort), demote `current` to `previous`, and mint a new `current`. Keeping the demoted token alive until the *next* refresh is the grace period: siblings still holding the old secret keep working for roughly one refresh interval.
- Regenerate the API token client from the updated OpenAPI spec (`by-snippet` lookup, create, delete, list).

## Companion PRs

This SDK release is a **prerequisite** for turning on finite identity Token TTLs in production via the ApiIdentity Reef work:
- Draft (reef): [axon-groundlight/zuuul#6651](https://github.com/axon-groundlight/zuuul/pull/6651) — identity-backed API Tokens management UI
- Draft (janzu): [axon-groundlight/zuuul#6650](https://github.com/axon-groundlight/zuuul/pull/6650) — device-api for ApiIdentity list / create / revoke

Those PRs are intentionally still drafts. They are substantially complete, but we will keep testing them and only merge after this SDK version is live, so customers (and Edge/hubs) that start receiving rotating tokens can refresh safely.

**Please review this PR together with #6650 and #6651** so the client rotation behavior and the identity management surface stay aligned.

Related backend hygiene: [axon-groundlight/zuuul#6681](https://github.com/axon-groundlight/zuuul/pull/6681) nulls out legacy backfilled `token_ttl` values so existing never-expire tokens stay static until users opt in.

## Rotation policy (client)

- Rotate iff identity `token_ttl` is non-null (from by-snippet / mint responses). No environment-variable override.
- Refresh cadence = `token_ttl / 30` (e.g. 30-day Token TTL => daily refresh).
- Per-token `expires_at` is still used to decide whether a cached secret is still usable, but not for cadence. Hard-deadline clamps therefore cannot accelerate refresh.
- Identity policy is applied by the server on mint; the client only reads `token_ttl` and `expires_at`.

## Dependency

- Requires public `/v1/api-tokens` `token_ttl` on list / create / by-snippet ([zuuul#6652](https://github.com/axon-groundlight/zuuul/pull/6652), **merged**; live on integ/eksdev, not yet on prod as of 2026-07-24).

## Test plan

- [x] Unit tests for null `token_ttl` (no rotate), hard-deadline-only (no rotate), cadence from `token_ttl` (not clamped `expires_at`), previous-parking, cleanup, and non-positive `token_ttl` clamp
- [x] Generated API token client coverage including `token_ttl`
- [x] Live check against api.dev with a short identity Token TTL (observed rotation with `token_ttl_seconds=180`, cadence ~6s)
- [ ] CI green on latest commit (currently blocked: prod docs tests fail because prod responses omit `token_ttl`)


tomfaulhaber (approved):
There's a lot here, but it looks pretty reasonable.

PR #466, opened 2026-07-29
Allow disabling automatic API token rotation

timmarkhuff:
## Background
Follow-up to #445 before cutting `0.31.0`. 

The `edge-endpoint` container on Edge Endpoint builds `Groundlight(api_token=request x-api-token)` for sync cloud calls. With rotation enabled it would mint/revoke against tokens that are owned by the client application. This could result in a token being revoked while the client is still using it. 

Disabling rotation here leaves ownership of the token's rotation with the application client.

## Summary
- Adds `enable_token_rotation` (default `True`) to `Groundlight` and `ExperimentalApi`.
- When `False`, `Groundlight` skips constructing `TokenManager` and uses the configured token as-is (no by-snippet lookup, on-disk cache, mint/revoke, or background refresh thread).
- `TokenManager` itself is unchanged from #445: it always owns a rotation chain when constructed.

## Test plan
- [x] `poetry run pytest test/unit/test_token_manager.py test/unit/test_token_refresh_client.py`
- [ ] Confirm default `enable_token_rotation=True` still rotates when `token_ttl` is set
- [ ] Edge follow-up: construct request-token clients with `enable_token_rotation=False`; keep rotation on for device-token containers with shared `GROUNDLIGHT_TOKEN_DIR`

### DIFF
--- a/src/groundlight/client.py
+++ b/src/groundlight/client.py
@@ -138,7 +138,7 @@ class Groundlight:  # pylint: disable=too-many-instance-attributes,too-many-publ
     POLLING_EXPONENTIAL_BACKOFF = 1.3  # This still has the nice backoff property that the max number of requests
     # is O(log(time)), but with 1.3 the guarantee is that the call will return no more than 30% late
 
-    def __init__(
+    def __init__(  # noqa: PLR0913  # pylint: disable=too-many-arguments
         self,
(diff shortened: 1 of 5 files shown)


honeytung (approved):
LGTM!

